PR #357, opened 2015-08-26
Fix deprecation warning of raise_exception

ayalash:


vmalloc:
Oops! Forgot about that one! Thanks!

PR #360, opened 2015-09-03
[skip_test] add condition for slash.skipped

eplaut:
This PR allows to set a slash.skipped decorator which applies only for a specific condition

### DIFF
--- a/slash/utils/__init__.py
+++ b/slash/utils/__init__.py
@@ -31,7 +31,7 @@ def skipped(thing=NOTHING, reason=None, condition=True):
     if isinstance(thing, str):
         return functools.partial(skipped, reason=thing, condition=condition)
     if isinstance(thing, type) and issubclass(thing, Test):
-        thing.skip_all(reason)
+        thing.skip_all(reason)  # pylint: disable=no-member
         return thing
 
     @functools.wraps(thing)
--- a/slash/utils/__init__.py
+++ b/slash/utils/__init__.py
@@ -30,15 +30,18 @@ def skipped(thing=NOTHING, reason=None, condition=True):
         return functools.partial(skipped, reason=reason, condition=condition)
     if isinstance(thing, str):
         return functools.partial(skipped, reason=thing, condition=condition)
-    if isinstance(thing, type) and issubclass(thing, Test):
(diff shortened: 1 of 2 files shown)


eplaut:
apparently, `tests/test_skips.py:test_method_with_condition_false` doesn't really checks the test ran.
Can you guide me how to change `tests/test_skips.py:_assert_not_skips`?

slash/utils/__init__.py
```diff
@@ -19,23 +20,29 @@ def repeat(num_repetitions):
...
 
 
-def skipped(thing, reason=None):
+def skipped(thing=NOTHING, reason=None, condition=True):
     """
     A decorator for skipping methods and classes
     """
     from ..core.test import Test
 
+    if thing == NOTHING:
```
vmalloc: When comparing sentinels, always use `is`, rather than `==`...

vmalloc:
@eplaut What's the difference between this and [test requirements](http://slash.readthedocs.org/en/master/tour.html#requirements)?

eplaut:
My shame for not understands that `slash.requires` is what I need.

Many thanks!

vmalloc:
It's all good :-) If you need anything else don't hesitate to ask

eplaut:
@vmalloc is there a way to set bunch of tests with a `slash.tag` and to make them run after other test (or another tagged tests) passed?

vmalloc:
Not currently, except for two consecutive runs. What's the use case?

eplaut:
I have bunch of test, some defined as sanity/smoke the others should not run if the sanity tests failed.

right now I change the test name to test_00_sanity so it run first and it set a `g` variable if it pass, and all the other has `@slash.requires` for that variable.

It cause 2 problems:
* The way to sort test is not cleared (and I'm not really sure it will work for any test name)
* It works for one sanity test, it will be more complicated for many of them (possible, but ugly)

vmalloc:
yeah, I understand. This could, in theory, be solved by a hook that can reorder tests prior to executions, based on tags. Currently I don't see any convenient way to solve this.

vmalloc:
Opened https://github.com/slash-testing/slash/issues/362

eplaut:
Thanks!

PR #382, opened 2015-11-01
[xunit-plugin] add test metadata

eplaut:
Add option to add extra data to the xunit xml.
```
xunit_plugin.add_test_metadata(title='system-out', content=test_url)  # add url to jenkins test results```

### DIFF
--- a/slash/plugins/builtin/xunit.py
+++ b/slash/plugins/builtin/xunit.py
@@ -48,10 +48,16 @@ class Plugin(PluginInterface):
 
     def _add_error(self, errortype):
         exc_type, exc_value, exc_tb = exc_info = sys.exc_info()
+        self.add_test_metadata(errortype,
+                               dict(type=exc_type.__name__,
+                                    message=str(exc_value)),
+                               get_traceback_string(exc_info))
+
+def add_test_metadata(self, title, attributes={}, content=None):
         test_element = self._get_xunit_elements_list()[-1]
-        error_element = E(errortype, dict(type=exc_type.__name__, message=str(exc_value)))
-        error_element.text = get_traceback_string(exc_info)
-        test_element.append(error_element)
+        data_element = E(title, attributes)
+        data_element = content
+        test_element.append(data_element)
 
     def _get_test_case_element(self, test):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vmalloc** [comment]
The Travis build seems to be failing on this PR

**Comment by eplaut** [comment]
pylint made me add an if statement, so the code coverage down by one line :(

File: slash/plugins/builtin/xunit.py [hunk 1]
```diff
@@ -48,10 +48,18 @@ def error_added(self, result, error):
 
     def _add_error(self, errortype):
         exc_type, exc_value, exc_tb = exc_info = sys.exc_info()
+        self._add_test_metadata(errortype,
+                                dict(type=exc_type.__name__,
+                                     message=str(exc_value)),
+                                get_traceback_string(exc_info))
+
+    def _add_test_metadata(self, title, attributes=None, content=None):
```
vmalloc: This shouldn't start with an underscore, as it implies being a private method. You should use a regular method, specifying it is not registered on any hook like in http://slash.readthedocs.org/en/master/plugins.html#plugin-activation

slash/plugins/builtin/xunit.py
```diff
@@ -48,10 +48,18 @@ def error_added(self, result, error):
 
     def _add_error(self, errortype):
         exc_type, exc_value, exc_tb = exc_info = sys.exc_info()
+        self._add_test_metadata(errortype,
+                                dict(type=exc_type.__name__,
```
vmalloc: General style note: `{'type': exc_type.__name__, 'message': str(exc_value)}` is more widely accepted for dictionary initialization and is also slightly faster

slash/plugins/builtin/xunit.py
```diff
@@ -48,10 +48,18 @@ def error_added(self, result, error):
...
+                                     message=str(exc_value)),
+                                get_traceback_string(exc_info))
+
+    def _add_test_metadata(self, title, attributes=None, content=None):
+        if not attributes:
+            attributes = {}
```
vmalloc: To prevent reduced coverage, and also to test this feature, a test is needed. If you are not comfortable enough writing one I'll add it later.

vmalloc:
Closed - implemented in 81a8f4b6a514d9add66df6014d913a07e34a3329 (#385)

PR #404, opened 2016-01-08
Add --with-tags option to slash_list

ayalash:


slash/frontend/slash_list.py
```diff
@@ -24,6 +24,7 @@ def _get_parser():
     parser.add_argument('-f', '--suite-file', dest='suite_files', action='append', default=[])
     parser.add_argument('--only-fixtures', dest='only', action='store_const', const='fixtures', default=None)
     parser.add_argument('--only-tests', dest='only', action='store_const', const='tests', default=None)
+    parser.add_argument('--with-tags', dest='with_tags', action='store_true', default=False)
```
vmalloc: Can we change to `--tags` or `--show-tags` to avoid confusion with plugins?

ayalash: Of course, my bad... I haven't thought about the plugins.

PR #411, opened 2016-02-10
[xunit-plugin] fix "Failed to add error to result" error

eplaut:
Hi,

When getting an error outside the test (like import error), using xunit plugin add another error of "Failed to add error to result" which hide the real error.
The PR is suggestion of work-around about the issue

Thanks,
Eli


vmalloc:
Thanks!

Issue #430, opened 2016-04-10
Add coverage plugin


eplaut:
It would be wonderful to have coverage report for test session.

vmalloc:
Sounds like a great idea! PRs are more than welcome!

vmalloc:
@eplaut I added the missing arguments to the proposed branch. Would appreciate if you could take it to a spin and let me know what needs improving.

I read the reference for coverage.py and am pretty confident adding multi-process support would be easier than the complexity in pytest-cov.

eplaut:
Many thanks, it looks great.

One issue, I think because of the differences between `activate` and `before_session_start` methods the report look a little bit strange, it misses the imports and function declaration.

vmalloc:
@eplaut thanks for the feedback! I moved the collection start to `activate` to fix the issue. Please let me know if it works for you now.

eplaut:
@vmalloc I've tested `feature/coverage` branch and it looks great. please put it in the next version.

Many thanks!

vmalloc:
Sure, no problem!

eplaut:
Hi @vmalloc, Can you add the coverage plugin to the current version (0.20)?

Many thanks!

vmalloc:
@eplaut unfortunately no, as we don't add new features to bugfix releases. You can try running with the develop branch, which will soon become v1.0, and has lots of other improvements as well

eplaut:
@vmalloc, I tried using develop, but it looks like `session_end` method doesn't run (and there is not report)

vmalloc:
@eplaut Can you provide something that reproduces the issue? I don't see such a behavior with the current version....

eplaut:
I'm trying to write some library, and there are no `.coverage` or `htmlcov` in the folder.
I've tried to debug it, and didn't seen any indication that the method ran (prints etc.)

vmalloc:
@eplaut `--with-coverage`?

eplaut:
yes. the activate method run for sure, but not the `session_end`

eplaut:
add `slash.plugins.manager.get_plugin('coverage').session_end()` in the end of the test does work

vmalloc:
@eplaut no that's not how it should be done. There's probably a bug somewhere. Will take a look.

eplaut:
I'm sure, just checked that the environment is good.
If you need me to debug, just tell me what to do.

vmalloc:
Found it. Opening linked issue

[...]

PR #431, opened 2016-04-10
[coverage-plugin] a working copy-paste of pytest-cov \o/

eplaut:
issue #430 
copy and paste from pytest (and nose), and change the hooks and config.
2 TODOs:
1. run _start() method before session_start (since we loosing some coverage)
2. add unittests on plugin


eplaut:
Hi @vmalloc 

For some reason, I can't bring the coverage to the original percent.
you can see in https://coveralls.io/jobs/13516666/source_files/221294155 that session_start/session_end/test_start/test_end are not covered for some reason.

in https://coveralls.io/builds/5755597/source?filename=slash%2Fplugins%2Fcov_engine%2Fembed.py there is poor coverage, since it only run when the test is running under other process (probably xdist side effect), and I'm not comfortable with remove it.

Can you help me on that?

vmalloc:
@eplaut will take a look. I think we can just instruct coveralls to ignore the coverage.py paths since they're not interesting.

vmalloc:
@eplaut several questions about this PR:
1. Why is all that mechanism necessary? AFAIK coverage.py (https://coverage.readthedocs.org/en/coverage-4.0.3/api.html) supports in-process activation and supports all the configuration needed. Is there a reason you chose pytest-cov for a copy/paste?
2. I think coverage reporting should be an optional plugin (requiring `--with-coverage` to activate), so having the plugin activated by default is a bit problematic

vmalloc:
I pushed a proposed baseline for this feature in the https://github.com/getslash/slash/tree/feature/coverage branch. It doesn't rely on all the heavy machinery of pytest-cov, so should be easier to maintain... If you have any strong preference to why we should insist on using pytest-cov, then we should consider it, otherwise I'm always in favor of going for the easier implementation...

eplaut:
1. I'm not aware of the differences between `activate` and `before_session_start`, I had to start coverage as soon as I can to get initial loading coverage (such as imports or function definition).
2. I do want the config part of `--cov`, `--cov-report`, and `--cov-append`. 
3. pytest-cov heavy mechanism comes to overcome xdist plugin coverage problems. I hope in the future `slash` will have the ability to split test to some threads too.

vmalloc:
@eplaut I agree about the xdist support. This is not about *threads*, but about *processes*, as threads are supported by coverage.

I believe there is a better way of merging coverage data than the path taken in pytest-cov, and we'll explore that when we land parallelization support...

eplaut:
sound reasonable. 

Anyway, I still think `--cov`, `--cov-report` and `--cov-append` options are valuable

vmalloc:
@eplaut sure, I'll add them.

vmalloc:
Moved discussion to related issue (#430)

PR #470, opened 2016-08-28
Fake tracebacks above calling point of handling_exceptions() context

gleon99:


slash/utils/traceback_proxy.py
```diff
@@ -0,0 +1,169 @@
+import sys
```
vmalloc: Please add a docstring explaining why this hack is needed and what it does, even in brief

vmalloc:
@gleon99 we need to add at least a basic test that makes sure the frames above are captured

PR #472, opened 2016-08-31
Fix incorrect traceback values assignment in handling_exceptions()

gleon99:


vmalloc:
@gleon99 the builds fail (see above)

vmalloc:
thanks!

Issue #403, opened 2016-01-06
New command to list plugins


omergertel:
Create new command to list plugins installed in the current slash environment:
`slash list-plugins`

Should return list of plugins - even better if you can tell if they are going to be activated.

This will help using the `--with` & `--without` flags

PR #480, opened 2016-10-18
#403 should also load local site for custom plugins

omergertel:


vmalloc:
Thanks for pointing it out. I will solve it with the app context which takes care of all the common environment stuff.

vmalloc:
Oh I'm stupid. Sorry. Reopening.

vmalloc:
This is still not complete. Will follow up with a fix for the "pending activation" mechanism

vmalloc:
Thanks!

PR #488, opened 2016-11-16
Fix documentation for slash.run_tests.

tobyk100:
Must load tests for their hooks to be registered, which is used by get_started_context to call the session start and end hooks.


vmalloc:
Thanks!

PR #489, opened 2016-11-20
Tagging: add get_tag function

eplaut:


slash/core/tagging.py
```diff
@@ -46,6 +46,9 @@ def __contains__(self, tag_name):
 
     has_tag = __contains__
 
+    def get_tag(self, tag_name, default_value=None):
+        return self._tags.get(tag_name, default_value)
```
vmalloc: default_value should be `NOTHING`, as this is how we denote a missing tag. `None` could be a legitimate tag value

vmalloc:
@eplaut I'm not sure about how this PR handles missing tags vs. empty-valued tags... What's the main use case of this method?

eplaut:
I tried to mimic the `dict.get` method which returns None for missing key. `NOTHING` was my first version of the function, but it is actually the default value for non-valued tag.

my use case uses tags to skip some tests by different argument relationship (so `-k` is not relevant) right now the code looks like this `name in tags`, `tag_value = tags[name]`then `my_arg in tag_value` . i wanted to make it nicer `my_arg in tags.get(name, [])`

p.s. i didn't check yet how `-k` interact with list value tag. soon it will be relevant...

vmalloc:
@eplaut so I think that it should raise a LookupError in case the default isn't provided (this can be easily achieved by a new sentinel, something like `NOT_GIVEN`). This will avoid possible confusion with `None` or `NOTHING`...

Issue #511, opened 2017-01-25
Add Result.get_log_paths()


ayalash:

ayalash:
@vmalloc I need some help with the changelog & cookbook section of this issue.
You can see my current branch (`ayalas/get_log_paths`) which is WIP.
Does my changelog message good enough?
What do you think I should add for the cookbook in this subject.

p.s.  I added docstrings for the Result's relevant methods, as we discussed, but I don't think they are good enough.

vmalloc:
@ayalash I think the cookbook should be something like:

`---------`
# Adding Multiple Log Files to the Current Result
Slash result objects contain the main path of the log file created by Slash (if logging is properly configured for the current run).

In some cases it may be desirable to include multiple log files for the current test. This can be useful, for example, if the current test runs additional tools or processes emitting additional logs:

`----------`

Then we can add an example of a simple test running an external command-line-based tool, giving it the generated log path with a `-L` argument or something

PR #514, opened 2017-01-30
Support additional log paths

ayalash:
Should fix https://github.com/getslash/slash/issues/511


doc/cookbook.rst
```diff
@@ -14,3 +14,23 @@ Slash offers a hook called ``tests_loaded`` which can be used, among else, to co
 	       test.__slash__.set_sort_key(index)
 
 The above code is best placed in a ``slashconf.py`` file at the root of your test repository.
+
+
+Adding Multiple Log Files to the Current Result
```
vmalloc: I think this would be more complete along with an example of how to add multiple log files to a single test as well..

PR #515, opened 2017-01-31
Add 'id' property to RunnableTest 

ayalash:
Should fix https://github.com/getslash/slash/issues/507

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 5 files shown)


vmalloc:
Changelog should be "Test id can now be obtained via `slash.context.test.id`"

Also, relevant lines that use workarounds should be changed - I see https://github.com/getslash/slash/blob/0c6a061450685fa1abefc62299e6fa9b5c29efb9/doc/logging.rst#L34, for example

ayalash:
@vmalloc I fixed my commit according to your comments. Please review again.

PR #518, opened 2017-02-06
Automatically detect yielded fixtures

ayalash:
Pull request for issue https://github.com/getslash/slash/issues/467

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`467` Automatically detect yielded fixtures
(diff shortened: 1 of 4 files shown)


doc/changelog.rst
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`467` Automatically detect yielded fixtures
```
vmalloc: "Slash now automatically detects yielded fixtures, and using `@yield_fixture` explicitly is no longer required"

vmalloc:
@ayalash the docs need to be updated as well

PR #522, opened 2017-02-12
Add '--no-output' flag for slash list (fix #519)

### LINKED ISSUE #519: Add quiet flag for slash list
This would allow checking everything can be imported without printing all the tests.

ayalash:


ayalash:
Related to issue https://github.com/getslash/slash/issues/519

slash/utils/cli_utils.py
```diff
@@ -167,6 +167,23 @@ def make_styler(style):
 UNDERLINED = '\x1b[4m'
 
 
+class Printer(object):
+
+    def __init__(self, report_stread, disable_output=False):
```
vmalloc: It's better to call this flag `enabled` and set it to `True` by default

slash/utils/cli_utils.py
```diff
@@ -167,6 +167,23 @@ def make_styler(style):
...
+        self._stream = report_stread
+        self._disabled = disable_output
+        self._func = self._no_op if disable_output else self._colored_print
+
+    def _no_op(self, *args):
```
vmalloc: This is an overkill IMO. It's better for `_colored_print` itself to check for enabled-ness

ayalash:
@vmalloc I fixed according to your comments. Please re-review.

PR #523, opened 2017-02-12
Fix unittests' pylint errors & run pylint on unittest in travis

ayalash:


ayalash:
@vmalloc I finally succeeded to pass travis.
Please review.
FYI (although I'm sure you already know) - When disabling pylint error inline of decorated function, python3.3's pylint expect the disabling comment to be on the decorator's line instead of the function's one.

tests/utils/suite_writer/slash_run_result.py
```diff
@@ -111,7 +111,7 @@ def __getitem__(self, args):
 
     def __contains__(self, args):
         return self._normalize_key(args) in self._events_by_args
-        return self.has_event(*args)
+        # return self.has_event(*args)
```
ayalash: @vmalloc According to git `self.had_event(*args)` is the newer code of this method, unfortunately, it fails on unittests.  Is it ok to keep the previous version?  or should we fix the `has_event` method?

tests/utils/suite_writer/test.py
```diff
@@ -80,13 +80,6 @@ def _write_epilogue(self, code_formatter):
         self._write_event(code_formatter, 'test_end')
         super(Test, self)._write_epilogue(code_formatter)
 
-    def _write_cleanups(self, code_formatter):
```
ayalash: @vmalloc I deleted this method because no one uses it & it uses `self._cleanups` which does not exist.

vmalloc: 👍

tests/utils/suite_writer/slash_run_result.py
vmalloc: @ayalash how does it fail exactly?

ayalash: It fails on rather (on not) an event exists. Assertions like:
```
assert not summary.events.has_event('NEVER')
assert evt1 in summary.events
assert cleanup in summary.events
```
(24 tests failed once I change this method to return `self.has_event(*args)`

vmalloc: @ayalash well has_event just calls `__contains__` again, so isn't it a Maximum recursion depth exceeded exception?

slash/core/fixtures/utils.py
```diff
@@ -26,7 +26,8 @@ def fixture(func=None, name=None, scope=None, autouse=False):
     return _ensure_fixture_info(func=func, name=name, scope=scope, autouse=autouse)
 
 nofixtures = function_marker('__slash_nofixtures__')
-nofixtures.__doc__ = 'Marks the decorated function as opting out of automatic fixture deduction. Slash will not attempt to parse needed fixtures from its argument list'
+nofixtures.__doc__ = 'Marks the decorated function as opting out of automatic fixture deduction. ' + \
```
vmalloc: No need for string concatenation. You can do
```
nofixtures.__doc__ = ("multiline "
"string")
```

vmalloc:
@ayalash this is some impressive piece of work. I'm surprised you survived it :-)

Thanks!

PR #526, opened 2017-02-14
Allow enabling/disabling output colors of list commands (fix #524)

### LINKED ISSUE #524: Allow enabling/disabling output colors of list commands
Currently, The output of `slash list`, `slash list-plugins` and `slash list-config` commands always uses colors.
It would be better to add `--color`/`--no-color` flags to those commands. The default (if none of the flags was given) should be color only if its tty.

ayalash:


slash/utils/cli_utils.py
```diff
@@ -128,24 +128,39 @@ def __init__(self, *args, **kwargs):
...
         self._stream = report_stread
         self._output_enabled = enable_output
+        self._should_colorize = None
+        self.set_colors(should_colorize)
 
-    def _colored_print(self, *args):
-        print(*args, file=self._stream)
+    def set_colors(self, should_colorize):
+        self._should_colorize = should_colorize if should_colorize is not None else self._stream.isatty()
+        Styler.should_colorize = self._should_colorize
```
ayalash: @vmalloc I don't like my solution for `should_colorize`, but it is the best I could find. If you have a better solution for it, I'll be happy to change.

slash/utils/cli_utils.py
```diff
@@ -128,24 +128,39 @@ def __init__(self, *args, **kwargs):
...
-    return lambda s: '{0}{1}{2}'.format(style, s, COLOR_RESET)
+    return Styler(style)
 
 UNDERLINED = '\x1b[4m'
 
 
 class Printer(object):
 
-    def __init__(self, report_stread, enable_output=True):
+    def __init__(self, report_stread, enable_output=True, should_colorize=None):
```
vmalloc: I would call it `force_color=False`

slash/utils/cli_utils.py
```diff
@@ -128,24 +128,39 @@ def __init__(self, *args, **kwargs):
...
         self._stream = report_stread
         self._output_enabled = enable_output
+        self._should_colorize = None
```
vmalloc: This should be `self._force_color = force_color`

slash/utils/cli_utils.py
```diff
@@ -128,24 +128,39 @@ def __init__(self, *args, **kwargs):
...
+        self.set_colors(should_colorize)
 
-    def _colored_print(self, *args):
-        print(*args, file=self._stream)
+    def set_colors(self, should_colorize):
```
vmalloc: There should be a method called `disable_coloring`, for disabling colors, and perhaps a method turning on force_color

slash/utils/cli_utils.py
vmalloc: See my above comments, and add to it a final method called `should_colorize` that takes everything into account:

```
should_color = self._force_color or (self._color_enabled and self._stream.isatty)
```

ayalash:
@vmalloc Please review again.

slash/frontend/list_config.py
```diff
@@ -16,13 +13,17 @@
 def _parse_args(args):
     parser = argparse.ArgumentParser(prog='slash list-config')
     parser.add_argument('paths', nargs='*', default=[])
+    parser.add_argument('--color', dest='wanted_color', action='store_const', const='color', default=None)
```
vmalloc: Why not `dest="force_color"` and `default=False` with `store_true`?

slash/frontend/list_config.py
```diff
@@ -16,13 +13,17 @@
...
+    parser.add_argument('--no-color', dest='wanted_color', action='store_const', const='bw', default=None)
```
vmalloc: Why not `store_false` and `dest="enable_clor"` with `default=True`?

slash/frontend/slash_list.py
```diff
@@ -44,7 +47,9 @@ def slash_list(args, report_stream=sys.stdout):
     if not parsed_args.paths and not parsed_args.suite_files:
         parser.error('Neither test paths nor suite files were specified')
 
-    _print = Printer(report_stream, enable_output=parsed_args.show_output)
+    force_color = parsed_args.wanted_color == 'color'
```
vmalloc: See comment above - this would make duplication here unnecessary...

slash/frontend/slash_list_plugins.py
```diff
@@ -19,14 +16,18 @@
...
 
 
 def slash_list_plugins(args, report_stream=sys.stdout):
-    _print = partial(print, file=report_stream)
-
     parser = _get_parser()
-    _ = parser.parse_args(args)
+    parsed_args = parser.parse_args(args)
+
+    force_color = parsed_args.wanted_color == 'color'
```
vmalloc: Same

ayalash:
@vmalloc Please review

vmalloc:
@ayalash thanks!

PR #531, opened 2017-03-01
Resume enhancements

maormarcus:


slash/core/variation.py
```diff
@@ -36,6 +37,9 @@ def _get_safe_repr(self):
                 returned[name] = '{}{}'.format(name, self.id[name])
         return ','.join('{}={}'.format(key, returned[key]) for key in sorted(returned))
 
+    def _get_resume_dict(self):
```
vmalloc: I think this should just use `id`...

slash/loader.py
```diff
@@ -108,8 +109,9 @@ def _address_in_file_matches(self, address_in_file, test):
         test_address_in_file = test.__slash__.address_in_file
         if address_in_file == test_address_in_file:
             return True
-        if '(' in test_address_in_file:
-            if address_in_file == test_address_in_file[:test_address_in_file.index('(')]:
+        if '{' in address_in_file and test_address_in_file == address_in_file[:address_in_file.index('{')]:
```
vmalloc: This seems like a mistake - Slash should not emit such test addresses...

slash/loader.py
```diff
@@ -108,8 +109,9 @@ def _address_in_file_matches(self, address_in_file, test):
         test_address_in_file = test.__slash__.address_in_file
```
vmalloc: I'm not sure you need this. You will get a variation with a proper id here, and you can compare against your saved id... Why do you need to parse the variation from the string?

slash/resuming.py
```diff
@@ -1,79 +1,95 @@
...
+    created_at = Column(DateTime, nullable=False, index=True)
+
+def init_db():
+    engine = create_engine('sqlite:///{0}/{1}'.format(_RESUME_DIR, _DB_NAME))
+    session.configure(bind=engine)
+    Base.metadata.create_all(engine)
+
+@contextmanager
+def connecting_to_db():
+    global is_db_initialized  # pylint: disable=global-statement
```
vmalloc: Call it `_db_initiatlized`. Otherwise:
1. It looks like a function
2. It looks public

slash/resuming.py
```diff
@@ -1,79 +1,95 @@
...
+    if not is_db_initialized:
+        init_db()
+        is_db_initialized = True
+    new_session = session()
+    try:
+        yield new_session
+        new_session.commit()
+    except:
```
vmalloc: ?

slash/resuming.py
```diff
@@ -1,79 +1,95 @@
...
+    finally:
+        new_session.close()
+
+def save_resume_state(session_result, collected):
+    metadata = SessionMetadata(
+                    session_id=session_result.session.id,
+                    src_folder=os.getcwd(),
+                    created_at=datetime.now())
+
+    passed_tests_names = [result.test_metadata.resume_repr for result in session_result.iter_test_results() if result.is_success_finished()]
```
vmalloc: I don't think we should rely on resume_repr

vmalloc: Also, I'm not sure resume_repr is properly sorted when generated. We should generate a field for variations which reliably stores and parses variation ids...

slash/resuming.py
vmalloc: There is no error handling here. In cases the db is locked by another session temporarily, we should gracefully retry...

slash/resuming.py
```diff
@@ -1,79 +1,95 @@
...
-        _logger.debug('Deleting old statefile {0!r}...', deleted_filename)
-        os.unlink(deleted_filename)
-
-def _get_resume_state_files_by_mtime():
-    return sorted(
-        ((os.stat(f).st_mtime, f) for f in
-         [os.path.join(_RESUME_DIR, f) for f in os.listdir(_RESUME_DIR)]),
-        reverse=True)
-
+    current_folder = os.getcwd()
```
vmalloc: This will, in many cases, be `"."`. You need to `os.path.abspath` it

tests/conftest.py
```diff
@@ -127,6 +127,7 @@ def fix_resume_path(request):
     # pylint: disable=protected-access
     prev = resuming._RESUME_DIR
     resuming._RESUME_DIR = tempfile.mkdtemp()
+    resuming.is_db_initialized = False
```
vmalloc: Please add a line asserting its existence prior to assignment. Otherwise it will break if (and when) you rename it

tests/test_slash_resume.py
```diff
@@ -1,54 +1,62 @@
...
+    def cleanup():  # pylint: disable=unused-variable
+        os.chdir(prev)
 
 def test_resume_no_session():
     with pytest.raises(CannotResume):
         get_tests_to_resume("nonexisting_session")
 
-
-def test_max_resumed_sessions(suite, patched_resume_dir):
-    for i in range(1, _MAX_NUM_SAVED_SESSIONS + 2):
```
vmalloc: I think this is still relevant. We don't want the resume db to bloat forever...

tests/test_slash_resume.py
```diff
@@ -1,54 +1,62 @@
...
-
     resumed = get_tests_to_resume(result.session.id)
 
     assert len(resumed) + result.session.results.get_num_started() - 1 == len(suite)
-    assert resumed[0].endswith(suite[fail_index].id)
+
+def test_resume_with_parametrization(suite):
+    num_values1 = 3
+    num_values2 = 5
+    test = suite.add_test(type='method')
```
vmalloc: This is already taken care of with the `suite_test` fixture

tests/test_slash_resume.py
```diff
@@ -1,54 +1,62 @@
...
+    test.add_parameter(num_values=num_values1)
+    test.add_parameter(num_values=num_values2)
+    fail_index = len(suite) // 2
+    suite[fail_index].when_run.fail()
+    summary = suite.run()
+
+    assert len(summary.get_all_results_for_test(test)) == num_values1 * num_values2
+    resumed = get_tests_to_resume(summary.session.id)
+    assert len(resumed) == 1
```
vmalloc: I think you should perform a second run and make sure the right test was run...

tests/test_slash_resume.py
```diff
@@ -1,54 +1,62 @@
...
+from slash.resuming import (CannotResume, get_last_resumeable_session_id, get_tests_to_resume)
 
-@pytest.fixture
-def patched_resume_dir(forge):
-    path = mkdtemp()
-    forge.replace_with(resuming, '_RESUME_DIR', path)
-    return path
+@pytest.fixture()
+def set_resume_cwd(request):
+    prev = os.getcwd()
```
vmalloc: abspath it

tests/test_slash_resume.py
```diff
@@ -1,54 +1,62 @@
...
+    assert len(resumed) == 1
+
+def test_different_folder_no_resume_session_id(suite, set_resume_cwd):  # pylint: disable=unused-argument
+    fail_index = len(suite) // 2
+    suite[fail_index].when_run.fail()
+    suite.run()
+    sessoin_id = get_last_resumeable_session_id()
+    assert sessoin_id
+
+    os.chdir(os.path.dirname(mkdtemp()))
```
vmalloc: you have the tmpdir fixture in pytest for this

vmalloc:
@maormarcus the Travis build is failing for this PR

PR #537, opened 2017-03-08
Report tests to backslash in advance

maormarcus:


vmalloc:
@maormarcus this still needs to be adjusted to use `tests_loaded`, right?

slash/frontend/slash_run.py
```diff
@@ -58,6 +56,14 @@ def slash_run(args, report_stream=None, resume=False, app_callback=None, working
 
 slash_resume = functools.partial(slash_run, resume=True)
 
+def report_tests_to_backslash(tests):
+    active_plugins = manager.get_active_plugins()
+    if 'backslash' not in active_plugins:
```
vmalloc: Better to `active_plugins.get('backslash', None)` and then compare to `None`. Prevents duplication

slash/frontend/slash_run.py
```diff
@@ -58,6 +56,14 @@ def slash_run(args, report_stream=None, resume=False, app_callback=None, working
...
+        return
+    backslash_plugin = active_plugins['backslash']
+    if hasattr(backslash_plugin, 'report_tests'):
+        backslash_plugin.report_tests(tests)
```
vmalloc: I think this should be called `report_planned_tests` to avoid confusion

PR #543, opened 2017-04-06
add option to set prefered debugger

alexshtein:


vmalloc:
Will be fixed in #544

PR #549, opened 2017-04-25
Fix provides/needs mechanism to handle "empty registrations" (fix #490)

### LINKED ISSUE #490: Fix plugin provides/needs scenarios
The following scenarios should work:

1. Plugin A provides X, at the class level, plugin B needs X, at the class level -- A must happen before B, even when some hooks/methods are missing from any of them
2. Plugin A provides X at the class level, plugin B has a specific hook that needs X -- all non-needing hooks should work, even when missing from A, the specific hook needs to happen in A before B. 
3. Same like 2, only A is missing the specific hook that B needs, should work exactly like number 2, only in the empty sense
4. B needs X at the class level, A only provides X on a specific hook - that in addition to B having other hooks, and A having other hooks. This case should fail, because logically the other hooks don't have anyone to provide X for them
5. B needs X at the class level, and has only one hook, that is provided for X by A as expected. This should work as expected. It will break once B adds more hooks.
6. same as 5, only inheriting more hooks from a parent plugin. This *does not* apply to `PluginInterfce`. In this case the child should inherit its parent's decorator. We should include an explanation about this in the docs.

ayalash:


ayalash:
@vmalloc Can we merge?

vmalloc:
Yup!

PR #561, opened 2017-05-16
Add warnings when finding duplicate test functions and…

maormarcus:
… fixtures

### DIFF
--- a/slash/loader.py
+++ b/slash/loader.py
@@ -23,6 +23,7 @@ from .exception_handling import handling_exceptions
 from .exceptions import CannotLoadTests
 from .core.runnable_test_factory import RunnableTestFactory
 from .utils.pattern_matching import Matcher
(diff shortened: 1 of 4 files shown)


slash/loader.py
```diff
@@ -51,10 +53,12 @@ def _get_matchers(self):
 
     def get_runnables(self, paths):
         assert context.session is not None
-
         sources = (t for repetition in range(config.root.run.repeat_all)
                    for t in self._generate_test_sources(paths))
         returned = self._collect(sources)
+        self._duplicate_funcs |= self._local_config.duplicate_funcs
+        for (path, name, line) in self._duplicate_funcs:
```
vmalloc: @maormarcus Perhaps it would be a good idea to use `sorted` to have the duplicates emitted in some sane order here

PR #563, opened 2017-05-21
Fallback to partial traceback if can't collect the full one

gleon99:


slash/exception_handling.py
```diff
@@ -83,7 +83,11 @@ def handling_exceptions(fake_traceback=True, **kwargs):
 
     if not PYPY and fake_traceback:
         # Only in CPython we're able to fake the original, full traceback
-        fake_tbs = create_traceback_proxy(frame_correction=2)
+        try:
+            fake_tbs = create_traceback_proxy(frame_correction=2)
+        except:
```
gleon99: Obviously, `except:` is far from being a good practice, but there's too many cases to handle when the underlying file had been changed (`IndexError` being not the only one..)

vmalloc: @gleon99 I'd at least narrow it down to `Exception`, to avoid `SystemExit`, and perhaps even further (should `KeyboardInterrupt` be swallowed?)

gleon99: @vmalloc I narrowed it down to only 2 types. Indeed better to narrow as much has possible.. I think these two shall be, probably,  enough. What do you say?

vmalloc:
Thanks!

PR #567, opened 2017-05-29
Fix issue with some windows file path

pierreluctg:
When using a test path like ```c:\path\to\my\test.py``` this was incorrectly detected as path="c" and address_in_file="\path\to\my\test.py"


vmalloc:
@pierreluctg Thanks a lot!

PR #569, opened 2017-05-30
Normalize the case of a path when traversing upwards

pierreluctg:
On windows this was causing issue if the drive letter specified by the user was lower case.


vmalloc:
Thanks for these fixes. We haven't had many Windows users so we're glad to get help with coverage and bug fixes there!

PR #570, opened 2017-05-30
Honor run.default_sources configuration when using slash list

pierreluctg:
When run.default_sources is defined in a configuration file, running ```slash list``` throws an error while ```slash run``` runs the tests in the run.default_sources.

This fix is to have ```slash list``` lists the same tests that will be executed by ```slash run``` when no test path or suite file are specified and run.default_sources is defined


vmalloc:
Thanks!

PR #587, opened 2017-06-18
Wait for all clients to connect before assigning tests

maormarcus:


slash/parallel/worker.py
```diff
@@ -49,12 +51,19 @@ def start(self):
...
             while not should_stop:
                 test_index = self.client.get_test(self.client_id)
-                if test_index == PROTOCOL_ERROR:
+                if test_index == WAITING_FOR_CLIENTS:
+                    _logger.debug("Worker_id {} recieved waiting_for_clients, sleeping".format(self.client_id))
+                    retries_num += 1
+                    if retries_num > WORKER_MAX_RETRIES:
+                        _logger.error("Worker_id {} waited max retries, disconnecting".format(self.client_id))
+                        break
+                    time.sleep(1)
```
vmalloc: I would change this to a smaller value. One second is a lot.

PR #602, opened 2017-06-25
Add `allow_raises` context (fix #595)

### LINKED ISSUE #595: Add allow_raises context

ayalash:


doc/changelog.rst
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`595` Add `allow_raises` context
```
vmalloc: ..., letting tests allow specific exceptions in selective context

doc/errors.rst
```diff
@@ -47,13 +47,30 @@ One case that is not easily covered by the assert statement is asserting Excepti
...
+
+.. code:: python
+
+	 >>> with slash.assert_raises(Exception) as caught: # doctest: +IGNORE_EXCEPTION_DETAIL
+	 ...    pass
+	 Traceback (most recent call last):
+	     ...
+	 TestFailed: ...
+
+In case the test should allow specific exception, but not enfore it (ex. timing issue), you can use :func:`slash.allow_raises`:
```
vmalloc:
```
In a case where the test author wants to allow a specific exception but not to enforce its propagation (e.g. allowing a timing issue to be present), :func:`slash.allow_raises` can be used.
```
Also add a note and a link explaining the relation of this to handling_exceptions

doc/errors.rst
```diff
@@ -47,13 +47,30 @@ One case that is not easily covered by the assert statement is asserting Excepti
...
+
+.. code:: python
+
+	  >>> with slash.allow_raises(Exception) as caught:
```
vmalloc: I'm having second thoughts about the name of this API. "Allow raises" doesn't make sense in english (should be "allow exception", or "allow to raise").

I'm not sure what the proper name for it should be -- maybe "allow_exception"? "allow_raising"? "might_raise"?

ayalash: I agree, I think that `allow_raising` sound the most like a context, what do you think?

doc/errors.rst
ayalash: About the note & link, There is this note:
https://github.com/getslash/slash/commit/68f90879d67892194d4452d27861c831f9fa91d7#diff-c430fdc89298c75a04be79e1cc7e2283R73
What should I add except for this?

doc/errors.rst
vmalloc: `allow_raising` still doesn't sound like a context, because it is not a progressive verb (it would be "allowing_raise" if we wanted this form).

Perhaps "allowing_exceptions"?

doc/errors.rst
vmalloc: Oh ok. It's fine then.

doc/errors.rst
ayalash: `allowing_excpetions` it is :)

ayalash:
@vmalloc I pushed a commit with the fixes, please review again.

vmalloc:
Thanks!

PR #608, opened 2017-06-26
Add blacklist to result's field names that shouldn't be pickled

maormarcus:


slash/core/result.py
```diff
@@ -25,6 +26,7 @@
 class Result(object):
     """Represents a single result for a test which was run
     """
+    pickle_key_blacklist = ['test_metadata', 'facts']
```
vmalloc: @maormarcus make this a `set` please:

`pickle_key_blacklist  = {'test_metadata', 'facts'}`

PR #610, opened 2017-06-27
Add hook before starting workers to allow adding them arguments

maormarcus:


slash/hooks.py
```diff
@@ -52,6 +52,7 @@ def _define(hook_name, **kwargs):
 
 _define('exception_caught_after_debugger',
         doc="Called whenever an exception is caught, and a debugger has already been run")
+_define('before_starting_workers', doc="Called before the parent starts the child workers", arg_names=("worker_args",))
```
vmalloc: 1. The form we use is "before_workers_start"
2. I think it is worth specifying that this is for parallel execution in the doc string
3. I think you should pass an object describing the worker to start, so that we can add more features to it in the future (like tweaking environment variables etc.)

PR #611, opened 2017-06-27
Add tmux features

maormarcus:


slash/conf.py
```diff
@@ -68,6 +68,7 @@
         "suite_files": [] // Doc("File(s) to be read for lists of tests to be run") // Cmdline(append="-f", metavar="FILENAME"),
         "stop_on_error": False // Doc("Stop execution when a test doesn't succeed") // Cmdline(on="-x"),
         "tmux": False // Doc("Run inside tmux") // Cmdline(on="--tmux"),
```
vmalloc: @maormarcus this is the point when it's worth moving it into a sub-config - `tmux` should become `tmux.enabled`, and the rest should be nested beside it...

PR #628, opened 2017-07-09
move server init outside try/except

eplaut:
Moving `server` initialization outside the try/except block removes the shadowing exception on the finally stage if server creation failed.

### DIFF
--- a/slash/parallel/server.py
+++ b/slash/parallel/server.py
@@ -185,8 +185,8 @@ class Server(object):
         return self.has_connected_clients() or self.has_more_tests()
 
     def serve(self):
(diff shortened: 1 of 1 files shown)


eplaut:
@vmalloc can you tell how mature the parallel feature is?
I'm willing to use it, but for some issues, it doesn't works for me. I'm having hard time to debug it and I want to read some documentation about it.

vmalloc:
@eplaut thanks a lot!

It's still in testing and we're still finding subtle bugs and nuances regarding how to work with it. We haven't written docs about it yet, but I'll add it to the list of things to get done before 1.3.0 is out

PR #636, opened 2017-07-12
Add documentation for parallel execution

maormarcus:


doc/parallel.rst
```diff
@@ -0,0 +1,49 @@
+.. _parallel:
+
+Parallel execution
+==============================
```
vmalloc: RST title lines should be at the exact length of the title

doc/parallel.rst
```diff
@@ -0,0 +1,49 @@
...
+
+By default, tests run in serial order. However, it is possible to run tests in parallel mode.
```
vmalloc: > A single Slash session process runs tests sequentially. However...

PR #652, opened 2017-07-31
Add host_fqdn & host_name attributes to session (fix #651)

### LINKED ISSUE #651: Attach hostname and fqdn to session object

ayalash:


slash/core/session.py
```diff
@@ -94,6 +95,8 @@ def deactivate(self):
 
     @contextmanager
     def get_started_context(self):
+        self.host_fqdn = socket.getfqdn()
```
vmalloc: @ayalash cache this value somewhere. When running tests with VPN enabled on a mac this can cause a 1000% slowdown

PR #653, opened 2017-08-04
Use IPython 5.x LTS for Python version before 3.3

pierreluctg:
Python 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
Beginning with IPython 6.0, Python 3.3 and above is required.


vmalloc:
Thanks! Much appreciated!

vmalloc:
Will release a hotfix with this change soon

PR #656, opened 2017-08-07
Improve email template

vmalloc:


slash/plugins/builtin/email_template.j2
```diff
@@ -0,0 +1,19 @@
...
+    <meta name="viewport" content="width=device-width"> <!-- Forcing initial-scale shouldn't be necessary -->
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"> <!-- Use the latest (edge) version of IE rendering engine -->
+    <meta name="x-apple-disable-message-reformatting">  <!-- Disable auto-scale in iOS 10 Mail entirely -->
+    <title>{{title}}</title> 
+</head>
+<body width="100%" bgcolor="silver">
+  <h1><img src="http://slash.readthedocs.org/en/latest/_static/slash-logo.png" style="height: 2.2em; margin-right: 1em; position: relative; top: 16px">{{title}}</h1>
+  <p>
+    Session <a href="{{backslash_link}}">{{session_id}}</a>, running on <strong>{{host_name}}</strong>  has finished {{result}} after {{duration}}.
+  </p>
```
ayalash: In case baskslash plugin is not active - there won't be `backslash_link`. 
Should we set it to empty string in `kwargs` or should we handle it here?

This bug is originated in previous commit which used `_HTML_TEMPLATE` (sorry - my mistake).

slash/plugins/builtin/notifications.py
```diff
@@ -200,8 +211,8 @@ def _get_message(self):
         else:
             session_info = 'Session ID: {session_id}'
         short_message = "{results_summary}\n\n" + session_info
-        title = "Slash session in {host_name} ended {result} (duration: {duration})"
```
vmalloc: @ayalash I thought about it too, but I'm not sure the duration is important enough for the subject. One can always just click the email to see the details...

slash/plugins/builtin/email_template.j2
vmalloc: It's ok - I tested it when there's no Backslash active and the result is an empty link (which is OK in HTML)

vmalloc:
@ayalash soon - I'm fixing the template

vmalloc:
Nevermind, will leave it like this for now

PR #666, opened 2017-08-27
Notifications improvements

ayalash:
@vmalloc This PR has 3 commits, one for each improvement.
I created a single PR, because they are all for notifications plugin.

### DIFF
--- a/slash/core/session.py
+++ b/slash/core/session.py
@@ -23,7 +23,7 @@ class Session(Activatable):
     """ Represents a slash session
     """
 
-    duration = start_time = end_time = host_fqdn = host_name = None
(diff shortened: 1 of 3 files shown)


slash/core/session.py
```diff
@@ -93,6 +93,12 @@ def deactivate(self):
         self._logging_context = None
         ctx.pop_context()
 
+    def get_current_duration(self):
```
vmalloc: @ayalash I prefer a single way of getting the duration. Perhaps make `duration` a property that will support both cases with the use of `start_time` and `end_time`?

ayalash:
@vmalloc I fixed the code according to your comments. Please review.

PR #695, opened 2017-10-25
Deprecate PluginInterface.get_config() and rename it to get_default_config() (fix #658)

### LINKED ISSUE #658: Deprecate PluginInterface.get_config(), rename to PluginInterface.get_default_config()

ayalash:


slash/plugins/__init__.py
```diff
@@ -213,6 +214,11 @@ def deactivate(self, plugin):
     def _configure(self, plugin):
         cfg = plugin.get_config()
         if cfg is not None:
+            warn_deprecation('PluginInterface.get_config() is deprecated since 1.5.0. '
```
vmalloc: @ayalash no need to specify the "since" version in text here. In normal deprecations this is only useful for documentation purposes, not in runtime

slash/plugins/interface.py
```diff
@@ -14,13 +14,19 @@ def activate(self):
...
+    def get_default_config(self):
         """
         Optional: should return a dictionary or a confetti object which will be placed under
         ``slash.config.plugin_config.<plugin_name>``
         """
         pass
 
+    def get_config(self):
+        """
+        Deprecated since 1.5.0. Please use PluginInterface.get_default_config() instead.
```
vmalloc: Please use http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated here

ayalash:
@vmalloc I pushed a commit with the changes, please review.

PR #715, opened 2018-01-09
Add --pdb-filter=FILTER command line argument (fix #712)

### LINKED ISSUE #712: Add "--pdb-filter=<filter_string>" command line argument
This argument will cause the session to enter PDB only if the filter_string is true (the filtering should be similar to "-k" argument).

ayalash:


doc/changelog.rst
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`712` Add ``--pdb-filter`` command line argument for filter the cases which runs debugger
```
vmalloc: `Added ``--pdb-filter`` - a new command-line flag that allows the user to enter pdb only on specific caught exceptions, based on pattern matching (similar to ``-k``)`

ayalash: Fixed

PR #727, opened 2018-02-04
Rename traceback_level to console_traceback_level (issue #702)

maormarcus:


doc/changelog.rst
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`702` Rename traceback_level to console_traceback_level
```
vmalloc: When naming configuration paths, please use the full path (`log.X`)

PR #728, opened 2018-02-06
Separate config for Slash's log levels from the global level filter (…

### LINKED ISSUE #719: Separate config for Slash's log levels from the global level filter
- [ ] Add extra parameters for log records requesting bypass of the separate filter (e.g. to cause tracebacks to be emitted to the user's log)

maormarcus:
…closes #719)

### DIFF
--- a/slash/conf.py
+++ b/slash/conf.py
@@ -45,6 +45,7 @@ config = Config({
                                                     decrease_doc='Make console more verbose (can be specified multiple times)',
                                                     increase='-q',
                                                     increase_doc='Make console less verbose (can be specified multiple times)'),
+        "core_log_level": logbook.WARNING // Doc("Minimal level of slash log messages to show"),
         "color_console": None // Cmdline(on='--force-color', off='--no-color'),
         "traceback_variables": False // Doc("Logs values of variables in traceback frames for added errors"),
         "console_traceback_level": 2 // Doc("Detail level of tracebacks") // Cmdline(arg="--tb"),
--- a/slash/log.py
+++ b/slash/log.py
@@ -18,7 +18,8 @@ from . import hooks
 _logger = logbook.Logger(__name__)
 
 _custom_colors = {}
-
+filtered_channels = {'slash.runner', 'slash.loader', 'slash.core.cleanup_manager', 'slash.core.scope_manager', \
+                      'slash.exception_handling', 'slash.core.fixtures.fixture_store'}
 
 class _NormalizedObject(object):
     def __init__(self, obj):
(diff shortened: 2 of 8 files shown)


slash/conf.py
```diff
@@ -45,6 +45,7 @@
                                                     decrease_doc='Make console more verbose (can be specified multiple times)',
                                                     increase='-q',
                                                     increase_doc='Make console less verbose (can be specified multiple times)'),
+        "filter_log_level": logbook.WARNING // Doc("Minimal level of slash log messages to show"),
```
vmalloc: I would call it `core_log_level`. `filter_log_level` is not a good name

slash/log.py
```diff
@@ -106,6 +106,11 @@ def emit(self, record):
             reporter.notify_after_console_output()
         return returned
 
+def _slash_logs_filter(record, handler): # pylint: disable=unused-argument
+    return record.extra.get('filter_bypass') or \
+           record.level >= config.root.log.filter_log_level or \
+           not record.channel.startswith('slash')
```
vmalloc: I think this should be a blacklist of channel names, not `startswith`. What if someone calls his or her logger `slashing` or `slash-infra`?

slash/log.py
```diff
@@ -106,6 +106,11 @@ def emit(self, record):
             reporter.notify_after_console_output()
         return returned
 
+def _slash_logs_filter(record, handler): # pylint: disable=unused-argument
+    return record.extra.get('filter_bypass') or \
+           record.level >= config.root.log.filter_log_level or \
```
vmalloc: the `or` ordering should be reversed IMO. set lookup for channel name is much faster than the mechanism implementing `__getattr__` for `config.root`

tests/test_logging.py
```diff
@@ -159,11 +159,35 @@ def on_session_start():
...
 
+@pytest.mark.parametrize('level', ['info', 'error'])
+@pytest.mark.parametrize('use_slash_logger', [True, False])
+def test_filtering_slash_logs(files_dir, suite, suite_test, config_override, level, use_slash_logger):
+    config_override('log.filter_log_level', logbook.WARNING)
+    assert slash.config.root.log.truncate_console_lines
 
+    long_string = 'a' * 1000
+    if use_slash_logger:
+        suite_test.append_line('slash.logger.{level}({msg!r})'.format(msg=long_string, level=level))
```
vmalloc: `slash.logger` is exactly an example of what *not* to filter, since it's advertised as a mechanism for the use of the test itself (See http://slash.readthedocs.io/en/master/tour.html#logging)

vmalloc:
@maormarcus need to add a changelog entry (something like `Added <config path>, allowing limiting the verbosity of logs initiated from Slash itself`)

slash/log.py
```diff
@@ -16,7 +16,8 @@
 _logger = logbook.Logger(__name__)
 
 _custom_colors = {}
-
+filtered_channels = ['slash.runner', 'slash.loader', 'slash.core.cleanup_manager', 'slash.core.scope_manager', \
```
vmalloc: Should be a set

maormarcus:
fixed

Issue #711, opened 2017-12-31
Support compressed logs


drrlvn:
There should be an option to have the logs written in a compressed format, for example using gzip. Logbook already supports this but it needs to be configured by slash.

ayalash:
Blocked by https://github.com/getlogbook/logbook/issues/251

PR #730, opened 2018-02-08
Support compressed logs (#711)

maormarcus:


slash/log.py
```diff
@@ -189,7 +199,7 @@ def _get_silenced_logs_context(self):
             return ExitStack()
         return SilencedLoggersHandler(config.root.log.silence_loggers).applicationbound()
 
-    def _get_file_log_handler(self, subpath, symlink, bubble=False, filter=None):
+    def _get_file_log_handler(self, subpath, symlink, bubble=False, filter=None, use_compression=False, cyclic=False):
```
vmalloc: The term is "rotating", not "cyclic". Cyclic means that old log *lines* get run over by new lines...

slash/log.py
```diff
@@ -200,16 +210,29 @@ def _get_file_log_handler(self, subpath, symlink, bubble=False, filter=None):
         else:
             log_path = self._normalize_path(os.path.join(root_path, _format_log_path(subpath)))
             ensure_containing_directory(log_path)
-            handler = self._get_file_handler_class()(log_path, bubble=bubble, filter=filter)
+            if use_compression:
+                log_path += ".gz" if config.root.log.compression.algorithm == "gzip" else ".brotli"
```
vmalloc: I think we should only allow the two options we support, and add a failure otherwise. In the current state of things, passing `algorithm="rotem"` will pick brotli...

maormarcus:
Fixed

doc/changelog.rst
```diff
@@ -2,6 +2,7 @@ Changelog
 =========
 
 * :feature:`-` ``-X`` can now be used to turn off stop-on-error behavior. Useful if you have it on by default through a configuration file
+* :feature:`711` Logs can now optionally be compressed on-the-fly through the "log.compression....." configuration parameter
```
vmalloc: I meant to write the actual configuration path here, not ellipsis XD

slash/log.py
```diff
@@ -200,16 +210,33 @@ def _get_file_log_handler(self, subpath, symlink, bubble=False, filter=None):
...
+        kwargs = {"bubble": bubble, "filter": filter}
+        if use_compression:
+            if config.root.log.compression.algorithm == "gzip":
+                handler = logbook.GZIPCompressionHandler
+                filename += ".gz"
+            elif config.root.log.compression.algorithm == "brotli":
+                handler = logbook.BrotliCompressionHandler
+                filename += ".brotli"
+            else:
+                self.session.results.global_result.add_failure("Unrecognized compression algorithm")
```
vmalloc: Does this fail immediately? Why not just through a RuntimeError?

maormarcus:
Fixed

PR #746, opened 2018-02-20
Issue 741

maormarcus:


slash/log.py
```diff
@@ -212,13 +207,30 @@ def _error_added_filter(record, handler): # pylint: disable=unused-argument
...
-        return handler.applicationbound()
+        return handler, log_path
 
     def _get_silenced_logs_context(self):
         if not config.root.log.silence_loggers:
             return ExitStack()
         return SilencedLoggersHandler(config.root.log.silence_loggers).applicationbound()
 
+    @contextmanager
+    def _wrap_with_cleanup(self, ctx, path):
```
vmalloc: I think it should be called `_deleting_files_if_needed(handler)` and deduce the path inside

vmalloc:
Thanks!

Issue #750, opened 2018-02-21
Log cleanup: FileNotFoundError trying to remove highlights


drrlvn:
```
Traceback (most recent call last):
  File "/home/drorl/dev/slash/slash/frontend/slash_run.py", line 74, in slash_run
    app.set_exit_code(-1)
  File "/home/drorl/dev/slash/slash/app.py", line 153, in __exit__
    self._exit_stack.__exit__(exc_type, exc_value, exc_tb)
  File "/usr/lib/python3.6/contextlib.py", line 380, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/drorl/dev/slash/slash/app.py", line 179, in _sigterm_context
    yield
  File "/usr/lib/python3.6/contextlib.py", line 365, in __exit__
    if cb(*exc_details):
  File "/usr/lib/python3.6/contextlib.py", line 284, in _exit_wrapper
    return cm_exit(cm, *exc_details)
  File "/home/drorl/dev/slash/slash/interfaces.py", line 10, in __exit__
    self.deactivate()
  File "/home/drorl/dev/slash/slash/core/session.py", line 91, in deactivate
    self._logging_context.__exit__(*exc_info) # pylint: disable=no-member
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/drorl/dev/slash/slash/log.py", line 160, in get_session_logging_context
    yield path
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/drorl/dev/slash/slash/log.py", line 188, in _get_file_logging_context
    yield handler, path
  File "/usr/lib/python3.6/contextlib.py", line 380, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/drorl/dev/slash/slash/log.py", line 226, in _file_handler_cleanup_context
    yield handler
  File "/usr/lib/python3.6/contextlib.py", line 365, in __exit__
    if cb(*exc_details):
  File "/usr/lib/python3.6/contextlib.py", line 284, in _exit_wrapper
    return cm_exit(cm, *exc_details)
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
[...]

PR #751, opened 2018-02-22
Refactor log cleanup contexts

maormarcus:
…ses #750)

### DIFF
--- a/slash/log.py
+++ b/slash/log.py
@@ -13,6 +13,7 @@ from .conf import config
 from .utils.path import ensure_containing_directory
 from .warnings import WarnHandler
 from .exceptions import InvalidConfiguraion
+from .exception_handling import handling_exceptions
 from . import hooks
 
 _logger = logbook.Logger(__name__)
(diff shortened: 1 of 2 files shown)


slash/log.py
```diff
@@ -193,85 +191,82 @@ def _should_delete_log(self, result):
...
+                    self._try_create_symlink(log_path, symlink)
+                with closing(self._create_log_file_handler(log_path, bubble=bubble, use_compression=use_compression, \
+                                                           use_rotation=use_rotation, filter=filter)) as handler:
+                    self._log_path_to_handler[log_path] = handler
+                    self._set_formatting(handler, config.root.log.format)
+                    yield handler
+                self._log_path_to_handler[log_path] = None
+                hooks.log_file_closed(path=log_path, result=result)  # pylint: disable=no-member
+                if config.root.log.cleanup.enabled and self._should_delete_log(result):
+                    os.remove(log_path)
```
vmalloc: Not entirely related to this PR, but I think failure to remove a log should just add an error to the session and not fail it completely...

PR #778, opened 2018-04-04
Remove old style string formatting from the code

ayalash:


slash/core/session.py
```diff
@@ -27,7 +27,7 @@ class Session(Activatable):
     def __init__(self, reporter=None, console_stream=None):
         super(Session, self).__init__()
         self.parent_session_id = config.root.parallel.parent_session_id
-        self.id = "{0}_0".format(uuid.uuid1()) if not self.parent_session_id else \
+        self.id = "{}_".format(uuid.uuid1()) if not self.parent_session_id else \
```
vmalloc: this seems like a wrong search/replace

ayalash: Fixed. Please review

vmalloc:
Thanks!

PR #787, opened 2018-04-16
Fix handling exceptions which raised from None in interactive session (closes #772)

### LINKED ISSUE #772: Interactive test does not handle correctly exceptions which raised from None
The following error is not detected by interactive test:
```
six.raise_from(Exception('Ahhh'), None)
```

ayalash:


vmalloc:
Thanks!

Issue #785, opened 2018-04-15
Support marker for plugins that are compatible with parallel mode


vmalloc:
Should be something like:

```python
@slash.plugins.parallel_mode('disabled')
class MyPlugin(PluginInterface):
   ...
```
Where "disabled" means disabling the plugin during parallel runs, "enabled" means enabling it, and also supporting "parent-only" and "child-only".

PR #790, opened 2018-04-22
Support marker for plugins that are compatible with parallel mode (cl…

maormarcus:
…oses #785)


slash/plugins/__init__.py
```diff
@@ -119,6 +121,23 @@ def is_internal_plugin(self, plugin):
         plugin_name = plugin if isinstance(plugin, str) else plugin.get_name()
         return self._installed[plugin_name].is_internal
 
+    def _is_parallel_deactivation_needed(self, plugin):
```
vmalloc: I think this should be named "is_parallel_supported", which better suits the usages in this file...

slash/utils/parallel_utils.py
```diff
@@ -0,0 +1,17 @@
+from ..conf import config
+
+def is_parallel_session():
+    return config.root.parallel.num_workers != 0
+def is_parent_session():
```
vmalloc: Add spaces between functions

slash/utils/parallel_utils.py
```diff
@@ -0,0 +1,17 @@
+from ..conf import config
+
+def is_parallel_session():
```
vmalloc: Add double space between functions

tests/test_plugin_parallel.py
```diff
@@ -0,0 +1,73 @@
...
+    if session_state == 'child':
+        assert parallel_utils.is_parallel_session()
+        assert parallel_utils.is_child_session()
+        assert not parallel_utils.is_parent_session()
+
+    plugin = mark('parallel_mode', plugin_parallel_mode)(plugin)
+    assert try_get_mark(plugin, 'parallel_mode') == plugin_parallel_mode
+
+    slash.plugins.manager.install(plugin)
+    slash.plugins.manager.deactivate_parallel_mode()
```
vmalloc: I think this should be renamed... Perhaps `configure_for_parallel_mode`?

PR #791, opened 2018-04-26
Notifications plugin should not do anything if it is running within a…

### LINKED ISSUE #776: Notifications plugin should not do anything if it is running within a parallel child

maormarcus:
… parallel child (closes #776)


slash/plugins/builtin/notifications.py
```diff
@@ -81,7 +82,7 @@ def get_html_message(self):
         return self._get_html_template().render(message=self, **self.kwargs)
 
 
-
+@parallel_mode(parallel_utils.ParallelPluginModes.PARENT_ONLY)
```
vmalloc: `@parallel_mode('parent-only')`

PR #794, opened 2018-05-01
Forward registers_on keyword arguments to gossip.register (closes #771)

### LINKED ISSUE #771: Make regsiters_on forward all unknown parameters to gossip.register

vmalloc:


slash/plugins/__init__.py
```diff
@@ -1,14 +1,19 @@
+from collections import namedtuple
```
ayalash: @vmalloc You added import to `namedtuple`, but you didn't used it

tests/test_plugins.py
```diff
@@ -28,6 +28,32 @@ def some_method_here(self):
...
+            pass
+
+
+    @gossip.register('some.hook', provides=['some_requirement', 'other_requirement'])
+    def _unused():
+        pass
+
+    gossip.trigger('some.hook')
+    hook = gossip.get_hook('some.hook')
+    [r] = [r for r in hook.get_registrations() if r.func.__name__ == 'plugin_method']
```
ayalash: Can we please change `r` to `reg` or something else? (`r` is return in pdb)

ayalash: Do we want to check the `registers_on` plugin with `plugins.provides` and `plugins.needs` in class level (of the plugin) works as well?

vmalloc:
@ayalash fixed and pushed. Please review again

PR #795, opened 2018-05-01
Adding support for details coming from result.details.append in xunit

pierreluctg:
Currently the xnuit plugin do not supports result details that are not
directly xml serializable. This includes details that can be added
using slash.context.result.details.append (detail value becomes a list).

With this change, slash.context.result.details.append is fully
supported in addition with supports for other detail values like dict,
list and tuple.


vmalloc:
@pierreluctg cool addition! Is there any chance you can add an accompanying unit test for this feature?

pierreluctg:
@vmalloc, unit tests added

vmalloc:
Thanks @pierreluctg! Will merge soon!

PR #796, opened 2018-05-02
Set 'Interactive' metadata for interactive test only (moved from Metdata.mark_interactive)

ayalash:


ayalash:
@vmalloc  I just found https://github.com/getslash/slash/issues/761 ("Metadata.mark_interactive() should not touch file path or address"), but it was marked for 1.7.0. 

Can my change be for this release (1.6.0)?
I'm changing my commit to marked 761 as closed by it.

PR #802, opened 2018-06-06
Support computed parameters (closes #672)

### LINKED ISSUE #672: Support dynamic parameters, computed only once the test using them actually starts

maormarcus:


vmalloc:
1. the code:

```
import itertools
import slash

_counter = itertools.count()

@slash.fixture
@slash.parametrize('param', [
    slash.param(compute=lambda: next(_counter), label='first'),
])
def fixture():
    pass

def test_something(fixture):
    pass
```
breaks.
2. This should work, and the repr of the test should be the same for both tests:

```
import itertools
import slash

_counter = itertools.count()

@slash.fixture(scope='module')
@slash.parametrize('param', [
    slash.param(compute=lambda: next(_counter)),
   slash.param(compute=lambda: next(_counter)),
])
def fixture():
    pass

def test_1(fixture):
    pass

def test_2(fixture):
    pass
```

PR #806, opened 2018-06-18
Add extract traceback failure to log (closes #805)

### LINKED ISSUE #805: Unable to investigate warning message
In this line: 
https://github.com/getslash/slash/blob/66a77c77d9f365c975fdac8a5671fdf3189c6783/slash/exception_handling.py#L89

Slash writes warning to screen, but there is no way to investigate it.

We should probably write the traceback to log file.

ayalash:


vmalloc:
Thanks!

PR #828, opened 2018-08-16
Fix validate_collection to be more generic

guytish:
When using @slash.hooks.tests_loaded.register to shuffle the tests order the == not working as expected.

### DIFF
--- a/slash/parallel/server.py
+++ b/slash/parallel/server.py
@@ -111,7 +111,7 @@ class Server(object):
 
     @server_func
     def validate_collection(self, client_id, client_collection):
-        if not self.collection == client_collection:
+        if set(self.collection) - set(client_collection):
             _logger.error("Client_id {} sent wrong collection", client_id)
             return False
         self.num_collections_validated += 1


vmalloc:
@guytish thanks for the PR! Can you include a unit test that breaks without this fix and works with it?

vmalloc:
It also appears like the tests are failing...

guytish:
Yes, saw it. Will fix - thanks!

vmalloc:
The proposed solution has a main drawback when discussing `--repeat`, in which case a test with a similar key would appear multiple times, and the `set()` call would cause all instances to collapse, possibly not detecting missing tests.

We will fix the issue by taking a slightly different approach, in which both parent and child will identify the planned tests by their unique ids, which is more correct and will solve the randomizing case as well.

I opened https://github.com/getslash/slash/issues/831 to handle this issue properly.

In the mean time thanks for reporting this!

guytish:
Thank you! Sorry for that..

vmalloc:
@guytish nothing to be sorry about!

PR #832, opened 2018-08-29
Refactor parent/child protocol of identifying tests (closes #831)

### LINKED ISSUE #831: Refactor parent/child protocol of identifying tests

maormarcus:


slash/parallel/server.py
```diff
@@ -56,7 +56,8 @@ def __init__(self, tests):
         for i in range(len(tests)):
             self.unstarted_tests.put(i)
         self.clients_last_communication_time = {}
-        self.collection = [[test.__slash__.file_path, test.__slash__.function_name, test.__slash__.variation.id] for test in self.tests]
+        self.collection = [[test.__slash__.file_path, test.__slash__.function_name, test.__slash__.variation.safe_repr] for test in self.tests]
```
vmalloc: `safe_repr` is lossy - it is not guaranteed to sort reliably. The numeric ids must be preserved. If the problem is comaprison, you can use frozenset or some other variant computed off the safe_repr...

slash/parallel/worker.py
```diff
@@ -3,6 +3,7 @@
 import os
 import pickle
```
vmalloc: Is there any test we can add to demonstrate the previous drawbacks we discussed for the old protocol? I'd be more comfortable if there were such a test included...

slash/parallel/worker.py
```diff
@@ -58,34 +60,36 @@ def start_execution(self, app, collected_tests):
             os.setsid()
         register(self.warning_added)
         register(self.error_added)
-        collection = [(test.__slash__.file_path, test.__slash__.function_name, test.__slash__.variation.id) for test in collected_tests]
-
-        if not self.client.validate_collection(self.client_id, collection):
+        collection = [(test.__slash__.file_path, test.__slash__.function_name, json.dumps(test.__slash__.variation.id)) for test in collected_tests]
+        if not self.client.validate_collection(self.client_id, sorted(collection)):
```
vmalloc: There's still a problem here... the sorting order actually depends on the construction order of the variation dict, which means we still have a problem. I think the sorting part should use the ordered dictionary item tuples...

slash/core/variation.py
```diff
@@ -78,6 +79,9 @@ def has_value_for_parameter(self, param_id):
     def get_param_value(self, param):
         return self._store.get_value(self, param)
 
+    def dump_variation_dict(self):
+        return json.dumps({k: v for k, v in sorted(self.id.items())})
```
vmalloc: Add comment explaining why this is needed.

PR #834, opened 2018-08-29
Implement usefixtures decorator (closes #648)

### LINKED ISSUE #648: Implement usefixture decorator
Similar to [pytest.mark.usefixtures](https://docs.pytest.org/en/latest/fixture.html#usefixtures)

maormarcus:


slash/core/fixtures/fixture_store.py
```diff
@@ -56,10 +56,14 @@ def iter_active_fixtures(self):
...
 
     def call_with_fixtures(self, test_func, namespace, trigger_test_start=False, trigger_test_end=False):
-
         if not nofixtures.is_marked(test_func):
             fixture_names = self.get_required_fixture_names(test_func)
             kwargs = self.get_fixture_dict(fixture_names, namespace)
+            used_fixtures_decorator_names = getattr(test_func, '__extrafixtures__', None)
+            if used_fixtures_decorator_names is not None:
+                intersection_fixture_names = set(fixture_names) & set(used_fixtures_decorator_names)
+                assert not intersection_fixture_names, "using twice fixtures {}".format(intersection_fixture_names)
```
vmalloc: I'm not sure this is an actual error. Why is it?

slash/__init__.py
```diff
@@ -37,7 +37,7 @@
 from .core.exclusions import exclude
```
vmalloc: Add changelog entry

slash/__init__.py
```diff
@@ -37,7 +37,7 @@
...
 from .core.fixtures import parametrize, parameters
```
vmalloc: Add documentation about use_fixtures

slash/core/fixtures/fixture_store.py
vmalloc: If it's an error it shouldn't be an assert IMO

slash/core/fixtures/fixture_store.py
```diff
@@ -249,7 +253,7 @@ def get_fixture_by_argument(self, arg):
     def get_fixture_by_id(self, fixture_id):
         return self._fixtures_by_id[fixture_id]
 
-    def get_fixture_dict(self, fixture_names, namespace=None, get_values=True, skip_names=frozenset()):
+    def get_fixture_dict(self, fixture_names, namespace=None, get_values=True, skip_names=frozenset(), add_to_dict=True):
```
vmalloc: Call this new parameter "add_as_keyword_param", since this is what it means?

vmalloc: Perhaps the interface would be better as an extra argument of "extra_initialized_fixtures" or something...

tests/test_fixture_use.py
```diff
@@ -0,0 +1,27 @@
...
+    suite.run()
+
+def test_use_fixture_and_depend_on_it_not_allowd(suite, suite_test):
+    fixture = suite_test.file.add_fixture()
+    suite_test.depend_on_fixture(fixture)
+    suite_test.add_decorator('slash.use_fixtures({})'.format([fixture.name]))
+    suite_test.expect_failure()
+    summary = suite.run()
+    [result] = summary.get_all_results_for_test(suite_test)
+    assert "using twice fixtures" in result.get_failures()[0].message
```
vmalloc: Need to add a test making sure the cleanup of the extra fixtures gets called... this is a major concern in such a feature

slash/core/fixtures/utils.py
```diff
@@ -26,6 +26,15 @@ def fixture(func=None, name=None, scope=None, autouse=False):
         return yield_fixture(func=func, name=name, scope=scope, autouse=autouse)
     return _ensure_fixture_info(func=func, name=name, scope=scope, autouse=autouse)
 
+def use_fixtures(fixture_names):
+
+    def decorator(func):
+        func.__extrafixtures__ = fixture_names
```
vmalloc: This doesn't support chaining, which might be a real use case in the future. For example:

```
def needs_something_initialized_globally(func):
    return use_fixtures('blap')(func)

@needs_something_initialized_globally
@use_fixtures('another')
def test_something():
    ....
```

PR #845, opened 2018-09-11
ignore_warnings: Support regex matching for filename (closes #844)

### LINKED ISSUE #844: ignore_warnings: Support regex matching for filename

ayalash:


slash/warnings.py
```diff
@@ -147,22 +147,29 @@ def __init__(self, category, filename, lineno, message):
         self.lineno = lineno
         self.message = message
 
+    @staticmethod
+    def _does_str_or_regex_matches(regex_or_str, warning_str):
```
vmalloc: Rename to `_pattern_matches`

slash/warnings.py
```diff
@@ -147,22 +147,29 @@ def __init__(self, category, filename, lineno, message):
...
+        if regex_or_str is None:
+            return False
+        if hasattr(regex_or_str, 'match'):
+            if regex_or_str.match(warning_str):
+                return True
+        elif regex_or_str == warning_str:
```
vmalloc: Compare first. Most of the time it will be a string, so the comparison is a quick short-circuit

PR #850, opened 2018-09-13
Add 'slash.app_quit' hook (closes #843)

### LINKED ISSUE #843: Failed session_start registrations cause plugins to not receive session_end notifications
We need to add a "quit" hook that is always called in the end of a slash run. Such a hook will enable us to clean up in cases where `session_start` wasn't even called yet, and plugins will be able to utilize it (e.g. Backslash reporting session end even on session-wide errors etc.)

ayalash:


doc/changelog.rst
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`843` Add ``slash.app_quit`` hook
```
vmalloc: `..., a hook that is called whenever the Slash process finishes. This is useful to allow certain plugins to perform important cleanups on session-wide errors`

PR #861, opened 2018-10-01
Assert that no internal errors happen during suite write runs

vmalloc:


slash/core/session.py
```diff
@@ -51,6 +52,12 @@ def __init__(self, reporter=None, console_stream=None):
 
         self._skip_exc_types = (exceptions.SkipTest,)
 
+    def notify_internal_error(self, error): # pylint: disable=unused-argument
```
ayalash: Why should `error` be passed here, if it is not used? Is it for future use?

vmalloc: Yes, although on second thought I'll drop it.

vmalloc:
@ayalash fixed and pushed

ayalash:
Cool :)  Merged

vmalloc:
Thanks!

PR #867, opened 2018-10-02
Fix handling of exceptions during _run_single_test() before test started (fixes #864)

### LINKED ISSUE #864: Slash session summary indicates sucessfully when requirement predicate raised excpetion
Reproduction scenario:
```
import slash

def _predicate():
    assert False

@slash.requires(_predicate)
def test_me():
    pass
```

ayalash:


ayalash:
@vmalloc Please review.

IMO the changelog entry I wrote is not that great, but I could not think of a better one - If you have one, please advice.

ayalash:
@vmalloc Currnetly (in Result class):
```
    def is_not_run(self):
        return not self._started

```
But there is a scenario (like in the added unittest) that the test hasn't started but it failed. Shouldn't we add check for `has_errors_or_failures` before?

PR #886, opened 2018-10-15
INFRADEV-11243: Support --unstarted-only and --failed-only flags for …

maormarcus:
…slash resume

### DIFF
--- a/slash/resuming.py
+++ b/slash/resuming.py
@@ -11,13 +11,13 @@ from slash.plugins import manager
 from ast import literal_eval
 from .utils.path import ensure_directory
 from .conf import config
+from .__version__ import __backslash_version__
 
 _DB_NAME = 'resume_state_v1.db'
 _logger = logbook.Logger(__name__)
 Base = declarative_base()
 _session = sessionmaker()
 _db_initialized = False
-
 
 class ResumeState(Base):
     __tablename__ = 'resume_state'
@@ -34,6 +34,12 @@ class SessionMetadata(Base):
     session_id = Column(String, primary_key=True)
     src_folder = Column(String, nullable=False, index=True)
     created_at = Column(DateTime, nullable=False, index=True)
+
+
+class ResumeTestStatus(object):
+    PLANNED = 'planned'
+    SUCCESS = 'success'
+    FAILED = 'failed'
 
 
 class ResumedTestData(object):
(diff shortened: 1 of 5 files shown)


slash/resuming.py
```diff
@@ -125,21 +125,29 @@ def get_last_resumeable_session_id():
...
                 _logger.debug('Found session {} locally, but no tests exist', session_id)
                 return []
             first_tests, other_tests = [], []
             for test in session_tests:
                 if not get_successful_tests and test.status == "success":
                     continue
-                new_entry = ResumedTestData(test.file_name, test.function_name)
-                new_entry.variation = literal_eval(test.variation) if test.variation else None
+                variation = literal_eval(test.variation) if test.variation else None
+                new_entry = ResumedTestData(test.file_name, test.function_name, variation, test.status)
```
vmalloc: It's time to use keyword arguments here

slash/resuming.py
```diff
@@ -148,19 +156,20 @@ def get_tests_from_previous_session(session_id, get_successful_tests=False):
...
+    unstarted_only, failed_only = config.root.resume.unstarted_only, config.root.resume.failed_only
     backslash_plugin = manager.get_active_plugins().get('backslash', None)
     if not backslash_plugin:
         raise CannotResume("Could not find backslash plugin")
     if not hasattr(backslash_plugin, 'is_session_exist') or not hasattr(backslash_plugin, 'get_tests_to_resume'):
         raise CannotResume("Backslash plugin doesn't support remote resuming")
     if not backslash_plugin.is_session_exist(session_id):
         raise CannotResume("Could not find resume data for session {}".format(session_id))
-    remote_tests = backslash_plugin.get_tests_to_resume(session_id, get_successful, unstarted_first, failed_first)
+    remote_tests = backslash_plugin.get_tests_to_resume(session_id, get_successful, unstarted_first, failed_first, unstarted_only, failed_only)
```
vmalloc: This should depend on the version of backslash-python we have installed to avoid breaking existing environments

slash/resuming.py
```diff
@@ -125,21 +125,29 @@ def get_last_resumeable_session_id():
...
+    unstarted_only, failed_only = config.root.resume.unstarted_only, config.root.resume.failed_only
     if unstarted_first and failed_first:
         raise CannotResume("Got both failed_first and planned_first, cannot choose")
+    if unstarted_only and failed_only:
+        raise CannotResume("Got both unstarted_only and failed_only, cannot resume")
     with connecting_to_db() as conn:
         # pylint: disable=no-member
         if conn.query(SessionMetadata).filter(SessionMetadata.session_id == session_id).first():
-            session_tests = conn.query(ResumeState).filter(ResumeState.session_id == session_id).all()
+            session_tests_query = conn.query(ResumeState).filter(ResumeState.session_id == session_id)
```
vmalloc: I think this code should be reworked at this point. Have a separate helper strategy class/function that yields all tests and their attributes as munches (started=True/False, success=True/False), and then a separate logic that prepares a list to run from them

PR #892, opened 2018-10-18
Fix Repr of _Cleanup for callbacks with no __module__

ayalash:


ayalash:
@vmalloc Travis finished successfully. Please merge.

PR #903, opened 2018-12-05
Allow parallel worker execution on Windows

pierreluctg:
Adding try/except for platform without os.getpgid and os.setsid() support


slash/parallel/worker.py
```diff
@@ -67,8 +67,13 @@ def _stop_keepalive_thread(self):
             self._watchdog_thread.join()
 
     def start_execution(self, app, collected_tests):
-        if os.getpid() != os.getpgid(0):
-            os.setsid()
+        try:
+            if os.getpid() != os.getpgid(0):
```
vmalloc: I strongly prefer an initial check for the existence of those in the beginning of the file and then check it here. e.g:
```python
_HAS_PID_PGID = platform.system() != 'Windows'
```

slash/parallel/worker.py
```diff
@@ -67,8 +67,13 @@ def _stop_keepalive_thread(self):
...
+                os.setsid()
+        except AttributeError:
+            _logger.debug(
+                "os.getpgid and os.setsid() are not available on your platform"
```
vmalloc: Change to `Skipping setsid for process since it is not available` or something like that

slash/parallel/worker.py
vmalloc: I strongly prefer an initial check for the existence of those in the beginning of the file and then check it here. e.g:
```python
_HAS_PID_PGID = platform.system() != 'Windows'
```

PR #912, opened 2019-01-07
Correctly call fixture test_start for each test

pierreluctg:
Currently if there is an exception raised from one of the fixture test_start, the test before and other tests method are not called. In addition, this fixture test_start will not be called again until all the fixtures test_start have been called once without exceptions. This leads to funky results...

```python
import slash


@slash.fixture(scope='session')
def fixture1(this):
    slash.logger.info("Hi from fixture 1")

    @this.test_start
    def at_test_start():
        slash.logger.info("on test start fixture 1")
        raise Exception("from fixture 1")
    return 1


@slash.fixture(scope='session')
def fixture2(this):
    slash.logger.info("Hi from fixture 2")

    @this.test_start
    def at_test_start():
        slash.logger.info("on test start fixture 2")
        raise AssertionError("from fixture 2")
    return 2

class Sample(slash.Test):

    def before(self, fixture1, fixture2):
        self.one = fixture1
        self.two = fixture2

    @slash.parametrize('i', range(5))
    def test_sample(self, i):
        slash.logger.info("Hello Sample {}".format(i))
        slash.logger.info(self.one)
        slash.logger.info(self.two)
```

### DIFF
--- a/slash/core/fixtures/fixture_store.py
+++ b/slash/core/fixtures/fixture_store.py
@@ -68,17 +68,18 @@ class FixtureStore(object):
         else:
             kwargs = {}
 
-        if trigger_test_start:
-            for fixture in self.iter_active_fixtures():
-                fixture.call_test_start()
-
         try:
-            return test_func(**kwargs)
-        finally:
-            if trigger_test_end:
+            if trigger_test_start:
                 for fixture in self.iter_active_fixtures():
-                    with handling_exceptions(swallow=True):
-                        fixture.call_test_end()
+                    fixture.call_test_start()
+        finally:
+            try:
+                return test_func(**kwargs)  # pylint: disable=lost-exception
+            finally:
+                if trigger_test_end:
+                    for fixture in self.iter_active_fixtures():
+                        with handling_exceptions(swallow=True):
+                            fixture.call_test_end()
 
     def get_required_fixture_names(self, test_func):
         """Returns a list of fixture names needed by test_func.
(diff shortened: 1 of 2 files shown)


vmalloc:
@pierreluctg Thanks! Can you please include a unit test to demonstrate the bug, thus verifying it is fixed?

pierreluctg:
@vmalloc, I have added tests.

Before the change the Slash test defined in the new unit tests was ending with:
`2 successful, 0 skipped, 2 failed, 2 erroneous`

With the fix the result is:
`0 successful, 0 skipped, 0 failed, 6 erroneous`

vmalloc:
@pierreluctg 0 successful?

pierreluctg:
@vmalloc, yes. The newly added unit tests are passing, but not the Slash tests in them.

Since the first fixture is adding an error (`ZeroDivisionError`) and the second one is adding failure (`AssertionError`) on test start, we expect each test to be `erroneous`.

pierreluctg:
@vmalloc, for some reason, Travis CI did failed to report the status, but everything is passing.

vmalloc:
@pierreluctg Nice! Will merge once the build finishes. Thanks a lot!

pierreluctg:
@vmalloc as far as I can tell the build did finish, but the status was not reported here.

vmalloc:
@pierreluctg indeed... must be a bad Travis day :-)
Merging

vmalloc:
Once again, thanks a lot!

vmalloc:
@pierreluctg Are you using class tests a lot? We have a plan to deprecate them at some point...

PR #915, opened 2019-01-08
Escaping special characters __file__ path

pierreluctg:
On windows failing to escape the `\` is causing failure to compile the regular expressions


vmalloc:
Doh! so obvious... Thanks for this!

vmalloc:
Thanks a lot!

PR #921, opened 2019-01-10
slash list --relative not working on Windows path

pierreluctg:


vmalloc:
@pierreluctg you're a rock star! Thanks!

PR #927, opened 2019-03-03
Change return-code of failed sessions (fixes #926)

### LINKED ISSUE #926: Consider changing failed session's return code
`git bisect run` using a process return code as indication for good/bad commits.
The problem is that it allow return codes between 0 to 128, but `slash run` returns - 1

ayalash:


vmalloc:
Thanks!

PR #929, opened 2019-03-08
Modifying the fixture test_start unit test to fix active fixture order

pierreluctg:
Previous implementation had parametrization on the fixture raising
ZeroDivisionError, this caused the second fixture called first staring
on the second tests.

This is because only the second fixture stay active.

I am not sure if we should fix this since we are not always honoring
the order of the fixture in the test method argument.


vmalloc:
In general fixture invocation order isn't guaranteed, and is only enforced by fixture dependency (which should always be upheld).

If a fixture raises, though, the remainder of "open" fixtures should be properly finalized...
Can you describe in more detail the scenario that isn't working properly here? I'm not sure I completely understand the issue...

pierreluctg:
See https://github.com/getslash/slash/pull/930

PR #930, opened 2019-03-08
Correctly propagating exceptions out of the test_start or test_end hooks

pierreluctg:
Restoring the behavior described in the **note** section here https://slash.readthedocs.io/en/master/fixtures.html#test-start-end-for-widely-scoped-fixtures

> Note: 
Exceptions propagating out of the test_start or test_end hooks will fail the test, possibly preventing it from starting properly

### DIFF
--- a/tests/test_fixture_start_end_test.py
+++ b/tests/test_fixture_start_end_test.py
@@ -60,34 +60,28 @@ def test_fixture_start_test_raises_exception(suite_builder):
         import slash  # pylint: disable=redefined-outer-name, reimported
 
         @slash.fixture(scope='module')
-        @slash.parametrize("param", [1, 2])
-        def fixture(this, param):
+        def fixture(this):
             @this.test_start
             def test_start(*_, **__):
                 1 / 0  # pylint: disable=pointless-statement
 
-            return param
+            return None
 
         @slash.fixture(scope='module')
-        def fixture_failing(this):
+        @slash.parametrize("param", [1, 2])
+        def fixture_failing(this, param):
             @this.test_start
             def test_start(*_, **__):
-                raise AssertionError()
+                slash.add_error("Not supposed to reach here").mark_fatal()
 
-            return None
+            return param
 
         class SomeTest(slash.Test):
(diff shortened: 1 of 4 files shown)


tests/test_fixture_start_end_test.py
```diff
@@ -97,34 +91,28 @@ def __code__():
...
         class SomeTest(slash.Test):
             def before(self, fixture, fixture_failing):  # pylint: disable=unused-argument,arguments-differ
-                self.data = fixture
+                self.data = 1
 
             @slash.parametrize("param", [10, 20, 30])
-            def test_something(self, param):
-                slash.context.result.data["value"] = param + self.data
-
-    suite_builder.build().run().assert_all(6).exception(ZeroDivisionError).with_data(
```
vmalloc: Why was this removed? I think it's important to make sure the correct values are passed to each iteration, as this is a very delicate mechanism

pierreluctg: I agree however in this case (with the fix in this PR) we are not supposed to execute the body of the test method when one of the the fixture `this.test_start` raises a exception. That is why the test method is now `slash.add_error("Not supposed to reach here").mark_fatal()`.

vmalloc: Oh I see now... That's right -- this behavior makes much more sense.

vmalloc:
@pierreluctg can you please add a changelog entry for this PR?

pierreluctg:
Hi @vmalloc, changelog entry added 🐛

vmalloc:
@pierreluctg thanks!

vmalloc:
Thanks for your contribution

pierreluctg:
Hi @vmalloc, can you please make maintenance/bugfix release with this change?

Thank you

vmalloc:
@pierreluctg oops! that's completely my bad -- I published a release branch but forgot to finalize the release. Will be done in the next few hours. Thanks for the reminder!

pierreluctg:
Thank you @vmalloc

PR #938, opened 2019-03-14
Refactor parallel to use worker_wrapper

maormarcus:


slash/parallel/parallel_manager.py
```diff
@@ -89,16 +58,8 @@ def start_server_in_thread(self, collected):
...
-            for worker_pid in self.server.worker_pids:
-                try:
-                    os.kill(worker_pid, signal.SIGTERM)
-                except OSError as err:
-                    if err.errno != errno.ESRCH:
-                        raise
-        else:
-            for worker in self.workers.values():
-                worker.send_signal(signal.SIGTERM)
+        for worker in self.workers.values():
```
vmalloc: I would turn that into a list in case at some point in the future the dictionary will change to modify itself based on worker availability:
```
for worker in list(self.workers.values()):
```

slash/parallel/parallel_manager.py
```diff
@@ -159,13 +116,9 @@ def check_no_requests_timeout(self):
...
     def start(self):
         self.try_connect()
-        if not config.root.parallel.server_port:
-            self.args.extend(['--parallel-port', str(self.server.port)])
-        if not config.root.parallel.keepalive_port:
-            self.args.extend(['--keepalive-port', str(self.keepalive_server.port)])
         try:
-            for _ in range(self.workers_num):
-                self.start_worker()
+            for worker_config in self.workers.values():
```
vmalloc: I would change that to `for worker in self.workers.values`

slash/parallel/tests_distributer.py
```diff
@@ -0,0 +1,46 @@
+from slash import ctx
+from logbook import Logger
+
+
+_logger = Logger(__name__)
+
+class TestsDistributer(object):
+    def __init__(self, tests):
```
vmalloc: Change to num_tests

slash/parallel/tests_distributer.py
```diff
@@ -0,0 +1,46 @@
...
+    def get_next_test_for_client(self, client_id):
+        ret = None
+        for test_index in self._unstarted_tests_indexes:
+            if self._can_execute_test(test_index, client_id):
+                ret = test_index
+                break
+            else:
+                _logger.debug('worker id {} cannot execute test number {}, searching another', client_id, test_index)
+        if ret is not None:
+            self._unstarted_tests_indexes.remove(ret)
```
vmalloc: Plural of `index` is `indices`

slash/parallel/worker_configuration.py
```diff
@@ -1,3 +1,137 @@
...
+                      '--workers-error-dir', ctx.session.parallel_manager.workers_error_dircetory]
+        self._start()
+
+    def kill(self):
+        raise NotImplementedError()
+
+    def handle_timeout(self):
+        raise NotImplementedError()
+
+    def exclude_test(self, test_index):
```
vmalloc: Add docstring to this, explaining semantics

slash/parallel/worker_configuration.py
```diff
@@ -1,3 +1,137 @@
...
+        if test_index in self._forced_tests:
+            raise RuntimeError('Cannot exclude test_index {} for client {} since it has been forced before'.format(test_index,
+                                                                                                                   self._worker_id))
+        self._excluded_tests.add(test_index)
+
+    def force_test(self, test_index):
```
vmalloc: Add docstring to this, explaining semantics

slash/parallel/worker_configuration.py
vmalloc: I would also make this method receive either a test object or a number (to make it easy for plugins/hook implementations)

slash/parallel/worker_configuration.py
vmalloc: Also validate input type here (see above)

slash/parallel/tests_distributer.py
vmalloc: ????

slash/parallel/worker_configuration.py
```diff
@@ -1,3 +1,156 @@
...
+                      '--workers-error-dir', ctx.session.parallel_manager.workers_error_dircetory]
+        self._start()
+
+    def kill(self):
+        raise NotImplementedError()
+
+    def handle_timeout(self):
+        raise NotImplementedError()
+
+    def _add_test_restriction(self, test_or_test_index, is_exclude):
```
vmalloc: This is actually two different functions disguised as one... why this structure?

slash/parallel/tests_distributer.py
maormarcus: sorry, forgot about it

PR #943, opened 2019-04-23
Add test execution time to xunit report

RonenHoffer:


slash/core/result.py
```diff
@@ -246,6 +250,13 @@ def add_skip(self, reason, append=True):
             self._skips.append(reason)
         context.reporter.report_test_skip_added(context.test, reason)
 
+    def get_execution_time(self):
```
vmalloc: I think a better term for this would be `duration`. Also, consider making this a `timedelta` object to avoid confusion - or change the method name to `get_duration_seconds` (I personally prefer the former)

vmalloc:
A couple of additional small things:

1. It is better to save the end time and then have duration computed from it, since it could, in theory, lose information
2. There is already a duration attribute on the global session object. It is probably better to move it to the session result and have it as an alias (I can do this as a followup change if you don't feel comfortable doing it)

RonenHoffer:
Hi,

I've did the following changes:
- `get_execution_time` changed to `get_duration` that returns a timedelta object
- The xunit content calls `timedelta`'s `total_seconds` function to get the execution time
- Add feature to changelog

And I have some questions:
- Should I amend the changes to the first commit?
- I've never used flux, can you point me for an example?

vmalloc:
1. You can push all the changes squashed into a single coherent commit
2. You can take a look at Backslash's tests doing this, e.g. https://github.com/getslash/backslash/blob/develop/tests/test_test_api.py#L61

RonenHoffer:
So I'm writing the tests in another project?
Because flux isn't part of this project and I'm not sure adding it as a dependency is really the right way.
WDYT @vmalloc ?

vmalloc:
@RonenHoffer that's a fair point. Let's skip the UT for now.

slash/core/result.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, test_metadata=None):
         self.test_metadata = test_metadata
         #: dictionary to be use by tests and plugins to store result-related information for later analysis
         self.data = {}
+        self._start_time = datetime.now()  # in case `mark_started` or `mark_finished` we're not called
+        self._end_time = self._start_time  # the computed value of duration should be 0
```
vmalloc: I think Duration should be `None` for non-started results, and for non-ended result should be either `None` or the actual time that elapsed.

slash/plugins/builtin/xunit.py
```diff
@@ -75,11 +75,13 @@ def session_end(self):
         if config.root.parallel.worker_id is not None:
             return
 
+        suite_time = sum([result.get_duration().total_seconds() for result in context.session.results.iter_test_results()])
```
vmalloc: timedelta objects support summing -
`sum((result.get_duration() for result in results), datetime.timedelta())`

RonenHoffer: done

slash/core/result.py
RonenHoffer: done

vmalloc:
Thanks for your contribution @RonenHoffer !

dyang36:
@vmalloc, when will there be a new release containing this feature?

vmalloc:
I guess we could schedule a release with it. @ayalash what do you think?

ayalash:
@vmalloc Sure, we'll release slash 1.8.0 this week

ayalash:
@dyang36 Slash 1.8.0 has just been released

dyang36:
Thank you all.

PR #947, opened 2019-05-19
add hook before parallel workers start and after they're all terminated

RonenHoffer:
Hi,

This PR comes after realizing that `session_start` and `session_end` hooks are invoked for each slash session that is started/ends when using the parallel worker.
Which means the code in the hook is run NUM_OF_PARALLEL_WORKES + 1 (for the main session).
I my use case I need to do setup before the workers start and teardown after they're done.
I think it's a nice feature to have, but if there's a solution I'm not aware of I would happily try it.

### DIFF
--- a/slash/hooks.py
+++ b/slash/hooks.py
@@ -68,6 +68,8 @@ _define('prepare_notification', doc='Called with a message object prior to it be
         arg_names=("message",))
 
 _define('before_interactive_shell', doc='Called before starting interactive shell', arg_names=("namespace",))
+_define('before_parallel_session_start', doc='Called before all parallel workers are initiated')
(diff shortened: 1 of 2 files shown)


vmalloc:
There are simpler, more elegant ways to solve this. For plugins, you can opt out of parallel runs (or certain flavors of it) using [the parallel mode toggle](https://slash.readthedocs.io/en/master/plugins.html#plugins-and-parallel-runs). For hooks, you can just check whether the session is a parallel parent through [the parent session id field](https://github.com/getslash/slash/blob/develop/slash/core/session.py#L30)...

RonenHoffer:
Thanks @vmalloc! that solved my issue, closing this PR

PR #960, opened 2019-06-25
Add support for tagging parameterization values

koreno:
How about this?
Not sure which style is most appropriate:
```
@slash.parametrize('x', [
    500 // slash.param(tags=["regression"]),
    5 // slash.param(tags="sanity"),
])
@slash.parametrize('y', [
    '100' // slash.tag("regression", "long"),
    '10' // slash.tag("regression", "short"),
    '1' // slash.tag("sanity"),
])
@slash.parametrize('z', [True, False])
def test_1(x, y, z):
    print(x, y, z)
```

There's one issue though when using tags with different values on different parameter-values, which generates tag conflicts.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,5 +1,6 @@
 Changelog
 =========
+* :feature:`-` Added support for tagging test parameterization values
 * :release:`1.8.0 <03-07-2019>`
 * :feature:`945` Drop support for deprecated arguments of ``add_cleanup``
 * :feature:`452` Drop support for old-style assertions
--- a/slash/core/fixtures/parameters.py
+++ b/slash/core/fixtures/parameters.py
@@ -8,6 +8,7 @@ from ..._compat import iteritems
 from ...exceptions import ParameterException
 from ...exception_handling import mark_exception_frame_correction
 from ...utils.python import wraps, get_argument_names
+from ..tagging import Tags
 from .fixture_base import FixtureBase
 from .utils import FixtureInfo, get_scope_by_name
 
@@ -147,6 +148,9 @@ class Parametrization(FixtureBase):
     def get_value_by_index(self, index):
         return self.transform(self._compute_value(self.values[index]))
 
+    def get_tags_by_index(self, index):
+        return self.values[index].tags
+
     def _compute_value(self, param):
         if isinstance(param, list):
             return [p.value for p in param]
@@ -174,11 +178,18 @@ class Parametrization(FixtureBase):
 
 class ParametrizationValue(object):
 
-    def __init__(self, label, value=NOTHING):
+    def __init__(self, label=NOTHING, value=NOTHING, tags=None):
         super(ParametrizationValue, self).__init__()
         self._validate_label(label)
         self.label = label
         self.value = value
+        self.tags = Tags()
+        if tags:
+            if not isinstance(tags, list):
+                tags = [tags]
+            for tag in tags:
+                name, value = tag if isinstance(tag, tuple) else (tag, NOTHING)
+                self.tags[name] = value
 
     def _validate_label(self, label):
         if isinstance(label, str) and not re.match(r'^[a-zA-Z_][0-9a-zA-Z_]{0,29}$', label):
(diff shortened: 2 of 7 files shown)


vmalloc:
I think the `tags=...` syntax looks better.

slash/core/fixtures/parameters.py
```diff
@@ -174,11 +178,18 @@ def as_transform(self, transform):
 
 class ParametrizationValue(object):
 
-    def __init__(self, label, value=NOTHING):
+    def __init__(self, label=NOTHING, value=NOTHING, tags=[]):
```
vmalloc: Avoid passing a list as a default parameter - it should default to either `None` or `()`

tests/test_slash_list.py
```diff
@@ -180,5 +180,42 @@ def test_slash_list_with_filtering(tmpdir):
     assert listed_tests == ['{}:{}'.format(fp.name, test_name) for test_name in ('test_a', 'test_c')]
 
 
+def test_slash_list_with_parameterization_filtering(tmpdir):
+    from textwrap import dedent
```
vmalloc: I just noticed that `textwrap` is not an official dependency, which means it could break in the future... It's not a problem with this PR, but we should fix it probably...

tests/test_slash_list.py
```diff
@@ -180,5 +180,42 @@ def test_slash_list_with_filtering(tmpdir):
...
+    with tmpdir.join('test_file.py').open('w') as fp:
+        _print = functools.partial(print, file=fp)
+        _print(dedent("""
+            import slash
+            @slash.parametrize('x', [
```
vmalloc: We have a better facility for generating and then asserting what runs -- take a look at https://github.com/getslash/slash/blob/a699ea65ea14d98d6cb5646bd29a0e6423cce2df/tests/test_parametrization_labels.py#L34 for an example on how it can be used

vmalloc:
This is an awesome addition! Thanks!
Some minor changes are needed and I'll merge it in

tests/test_slash_list.py
koreno: isn't it part of the standard lib?

vmalloc: Doh! I had it confused with `texttable`. Yes it is.

koreno:
> I think the `tags=...` syntax looks better.

when you get into tags with values it gets a little more obscure...:
`500 // slash.param(tags=[("regression", "long")])`

with `slash.tag` it remains consistent with the way you tag tests:

```
long_reg = slash.tag("regression", "long")

@long_reg
def test_1(): ...

@slash.parametrize('x', [500 // long_reg, 5])
def test_2(x): ...
```

tests/test_slash_list.py
koreno: I saw it, but wasn't sure how to specify `-k` filters to `suite_builder.build().run()`

vmalloc: It's ok to extend `run` to accept additional args. It doesn't accept them at the moment simply because it wasn't needed thus far.

vmalloc:
Let's start by supporting the simple case first. We can tweak it later in a separate PR (better not through checking callability, but through a dedicated ABC/base class)

vmalloc:
@koreno can you please rebase on top of develop?

doc/changelog.rst
```diff
@@ -2,6 +2,7 @@ Changelog
 =========
 * :release:`1.8.0 <03-07-2019>`
 * :feature:`945` Drop support for deprecated arguments of ``add_cleanup``
+* :feature:`-` Added support for tagging test parameterization values
```
vmalloc: @koreno this should be pushed up -- it's not a part of the 1.8 release

vmalloc:
Thanks!

koreno:
> 
> @slash.parametrize('x', [500 // long_reg, 5])
> def test_2(x): ...
> 
> @slash.parametrize('x', [long_reg(500), 5])  # we can make this work too, by checking for `callable`
> def test_3(x): ...
> ```

The `some_tag(val)` form becomes necessary when `val` already has a `__floordiv__` operator, and thus `val // some_tag` fails...

PR #963, opened 2019-07-03
Drop support for deprecated arguments of add_cleanup (closes #945)

### LINKED ISSUE #945: Drop support for deprecated arguments of add_cleanup
https://github.com/getslash/slash/issues/220 (added deprecation)[https://github.com/getslash/slash/blob/develop/slash/core/cleanup_manager.py#L66-L70] for non-explicit args/kwargs of add_cleanup:
```
if args or kwargs:
    warn_deprecation('Passing *args/**kwargs to slash.add_cleanup is deprecated. '
                     'Use args=(...) and/or kwargs={...} instead', frame_correction=+2)
```

This support should be dropped - Tests should fail if they uses this form.

ayalash:


ayalash:
@vmalloc The tests passed successfully - Can we merge it?

PR #965, opened 2019-08-02
Fix pattern matching for whitespaces after "tag:" (closes #964)

### LINKED ISSUE #964: '-k' has false results when there is space in the string
For example:
```
$ slash list tests -k 'tag:basic-io' --only-tests | wc -l
18
```
While (notice that space between `tag:` and `basic-io`
```
$ slash list tests -k 'tag: basic-io' --only-tests | wc -l
182
```

vmalloc:


ayalash:
@vmalloc Thanks!!

PR #976, opened 2019-09-19
support tagging parameterized tests using subscription syntax

koreno:
The current `//` syntax is limited to left-hand operands that don't _already_ implement a `__floordiv__` operator (and we have those...)

Here it is in use:
```
from easypy.units import MiB, GiB

@slash.parametrize('size', [
    tags.short[5 * MiB],
    tags.regression[GiB],
])
def test_write_and_read(size):
    ...
```


vmalloc:
Why square brackets? I think it is a bit confusing to read, and at that point I would prefer the `slash.param` syntax for such cases...

koreno:
> Why square brackets? I think it is a bit confusing to read, and at that point I would prefer the `slash.param` syntax for such cases...

The reason is that we've defined all "allowed" tags in one module, and we then decorate the tests with these predefined tags. It then makes more sense to reuse the tag objects on parameterized values as well:


```
import tags

@tags.regression
def test1():
   ....

@tags.sanity
def test2():
   ....

@slash.parametrize('size', [
            tags.sanity[5 * MiB],  # since MiB already defines `//`...
            tags.regression[500 * MiB],
        ])
def test3(size):
   ....

**Comment by vmalloc** [comment]
Ok, I'll try to explain why I'm not in favor of this suggestion, now that I have some time to write it down.

First, a word about our use of `//` for param labeling. I'm generally in favor of operators "reading like what they actually mean", but this is a bit of an exception -- very much like what we did in the `confetti` library. Of course we don't actually "divide" our parameters by their labels, but this operator *does* offer two readability advantages:

1. It resembles C/C++/Java/Rust/etc. comments -- meaning it reads like "the value is 20, and by the way its label is x"
2. It lets us put the parameter first, and its label afterward. Given labels often have wildly varying lengths as opposed to numerical (or size) values, this was advantageous.

Of course, `//` is only syntax sugar, so the "official" way of doing all this is with `slash.param`.

The subscript syntax doesn't offer any additional readability value, and is confusing -- does it mean accessing the label at the "value" index? Getting the value out of the label a-la-dictionary? Hard to tell for sure. 

Having said all that, keep in mind that it's super easy to create a tagger helper to achieve exactly what you want outside of Slash altogether, which is all the more reason not to diversify this syntax any further out of the box.

And as for the original reason why you claim you need this - you really shouldn't be needing it IMO, and it sounds like a bug in your DataSize library. For example, this works just fine with our `capacity` library:

```python
from capacity import MiB
import slash

@slash.parametrize("size", [
    MiB // slash.param("big")
])
def test_something(size):
    assert size > 0
```
I suspect the reason is that DataSize doesn't honor Python's operator overloading protocol with regards to `NotImplemented`, so fixing that would remove the need to have this hack implemented altogether.

koreno:
It does seem we missed a spot with the DataSize implementation...

I take your comments, but all things considered there is value in finding
ways of making these test "headers" as concise and clean as possible. The
fully keyworded function call buries the values in boilerplate, making it
hard for a human to parse. This makes the parameterization feature less
useful IMHO.
I'll take your suggestion though and move it into our infra instead.

ayalash:
Rejected due to previous comments

PR #980, opened 2019-10-16
Change OrderedSet package (closes #979)

### LINKED ISSUE #979: Change package of OrderedSet
Slash depend on [orderedset](https://pypi.org/project/orderedset/) package which uses deprecated API and wasn't updated for a long time.
There is a similar package [ordered-set](https://pypi.org/project/ordered-set/) which is maintanied and doesn't uses deprecated API.

ayalash:


doc/changelog.rst
```diff
@@ -1,5 +1,6 @@
 Changelog
 =========
+* :feature:`979` Change package of ``OrderedSet``
```
vmalloc: Change Ordered set implementation from `orderedset` to `ordered-set`, as the former is not maintained anymore

PR #982, opened 2019-10-23
Read file as binary

kbh2o:
This will avoid decoding errors when files are in different encodings


kbh2o:
sure thing

vmalloc:
Many thanks for your contribution!

PR #984, opened 2019-10-29
don't set attributes on threading.Event objects

koreno:
We use gevent's monkey patching which has 'read-only' version of these objects, and we fail:

```
Traceback (most recent call last):
  */slash/frontend/slash_run.py:64 ................ slash_run >> with app.session.get_started_context():                  
  /usr/local/lib/python3.7/contextlib.py:112 ...... __enter__ >> return next(self.gen)                                    
  */slash/core/session.py:122 ........... get_started_context >> with notify_if_slow_context("Initializing session..."):  
  /usr/local/lib/python3.7/contextlib.py:112 ...... __enter__ >> return next(self.gen)                                    
  */slash/utils/interactive.py:66 .... notify_if_slow_context >> evt.should_report_end_msg = False                        
AttributeError: 'gevent._event.Event' object has no attribute 'should_report_end_msg'
```

### DIFF
--- a/slash/utils/interactive.py
+++ b/slash/utils/interactive.py
@@ -63,11 +63,14 @@ def _humanize_time_delta(seconds):
 @contextmanager
 def notify_if_slow_context(message, slow_seconds=1, end_message=None, show_duration=True):
     evt = threading.Event()
-    evt.should_report_end_msg = False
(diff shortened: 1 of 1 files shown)


vmalloc:
Thanks!

PR #997, opened 2019-11-12
Convert float and integer test details to strings in xunit output files

bpowell65536:
This changeset adds conversion of float and integer test detail values to strings when writing xunit output, allowing tests which store float and integer test details to also generate xunit output.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 * :feature:`598` Drop support for python version < 3.6
 * :feature:`-` Added support for tagging test parameterization values
 * :feature:`-` Added a plugin which provides links to test logs which have been archived by a CI system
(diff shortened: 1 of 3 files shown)


doc/changelog.rst
```diff
@@ -8,6 +8,7 @@ Changelog
 * :feature:`598` Drop support for python version < 3.6
 * :feature:`-` Added support for tagging test parameterization values
 * :feature:`-` Added a plugin which provides links to test logs which have been archived by a CI system
+* :feature:`-` Convert float and integer test details to strings in xunit output files
```
vmalloc: I think this is actually a bugfix... phrasing should be that the behavior was corrected for test details containing floats in conjunction with xunit...

vmalloc:
@bpowell65536 thanks for your contribution!

PR #999, opened 2019-11-17
Choose IPython colors scheme for interactive test (fixes #998)

### LINKED ISSUE #998: Choose IPython color scheme
When running with IPython >= 7.2.0, slash interactive doesn't have colors anymore. See related IPython issue: ipython/ipython/issues/11523
According to that issue, the problem is solved by passing `colors` to `embed` function.

From IPython's documentation:
> It has four case-insensitive values: 'nocolor', 'neutral', 'linux', 'lightbg'. The default is neutral, which should be legible on either dark or light terminal backgrounds. linux is optimised for dark backgrounds and lightbg for light ones.

https://ipython.readthedocs.io/en/stable/config/details.html

ayalash:


vmalloc:
thanks!

PR #1000, opened 2019-11-21
support 'success_only_cleanup' yield fixture

koreno:
given success_only_cleanup=True, a yield_fixture will not call the 'exit' section
for the fixture if the test failed. this is useful when the cleanup can make it
difficult to troubleshoot/debug the failed test.

### DIFF
--- a/slash/core/fixtures/active_fixture.py
+++ b/slash/core/fixtures/active_fixture.py
@@ -1,4 +1,7 @@
+import functools
 from ...exception_handling import handling_exceptions
+from ...ctx import context as slash_context
+
 
 class ActiveFixture(object):
 
@@ -33,10 +36,19 @@ class ActiveFixture(object):
             for callback in self._test_end_callbacks:
                 callback()
 
-    def add_cleanup(self, cleanup):
-        self._cleanups.append(cleanup)
+    def add_cleanup(self, cleanup=None, success_only_cleanup=False):
+        if cleanup is None:
+            return functools.partial(self.add_cleanup, success_only_cleanup=success_only_cleanup)
+
+        self._cleanups.append((cleanup, success_only_cleanup))
 
     def do_cleanups(self):
+        if not slash_context.result:
+            # for when we're unit-testing, and there's not 'result'
+            success = True
+        else:
+            success = slash_context.result.is_success()
         while self._cleanups:
-            cleanup = self._cleanups.pop()
-            cleanup()
+            cleanup, success_only_cleanup = self._cleanups.pop()
+            if not success_only_cleanup or success:
+                cleanup()
--- a/slash/core/fixtures/utils.py
+++ b/slash/core/fixtures/utils.py
@@ -128,13 +128,15 @@ def yield_fixture(func=None, **kw):
     if func is None:
         return functools.partial(yield_fixture, **kw)
 
+    success_only_cleanup = kw.pop('success_only_cleanup', False)
+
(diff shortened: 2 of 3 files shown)


vmalloc:
Contexts are problematic for success-only cleanups, and it's basically impossible to guarantee success-only semantics for all possible cases. Let's do a little though experiment.

Let's say I want a fixture which sets up a directory of files, and deletes it at the end, using a utility "database" that was written before (don't stick to this specific example, it's symbolizes a whole class of fixture cases). I might have this:

```python
class DirectoryDatabase:
    def __enter__(self):
        print("Setting up...")

    def __exit__(self, *_):
        print("Destroying all evidence!!!")
```
Now, without going into the exact code in Slash that supports yield fixtures, we can paraphrase our fixture like this:
```python
def myfixture():
    with DirectoryDatabase() as db:
        yield SomeObject(db)
```
What you're trying to do is basically stripped down to:
```python
generator = myfixture()
next(generator)
```
Note the lack of the extra `next` - we're trying to avoid calling the cleanup, right? 
Well, in this particular case, you will allways see the "destroying evidence" message. The DB will always get cleaned up, even if failures occur, to the user's dismay...

Trying to `contextmanager` our way out of it, of course, doesn't help either:
```python
@contextlib.contextmanager
def myfixture():
    with DirectoryDatabase() as db:
        yield SomeObject(db)
...
g = myfixture()
g.__enter__() # no exit
```
But still, the database gets destroyed. This is of course due to GeneratorExit.
The same behavior goes if somewhere within our fixture (or through contexts or contexts of contexts) there's a use of `try/finally`, which is very very common...

koreno:
If I understand correctly, you're saying that when context-managers are garbage collected by python, it won't matter that slash didn't explicitly called `next` on those yield-fixtures, since python will indirectly call them for us by closing the generators and thereby raising a `GeneratorExit` inside. If the user used a `try/finally` in their context manager, the `finally` clause will be executed regardless of our desire for `success-only` behavior.

Well - 

1. I was unable to reproduce this in the unit-tests I wrote. Maybe you could demonstrate this unwanted behavior by modifying my unittest somehow?
2. tr/except, try/finally - these are useless within yield_fixtures since these yield fixtures never receive the test exceptions. It might be another solution to this problem - a `yield_fixture` that _does_ get the test exceptions thrown into it, so it could decide if/when to perform its cleanup. Of course it wouldn't be allowed to swallow the test failure, simply decide how to handle cleanups.
3. if the user uses a ctx-manager which has a try/finally - well then I'd say it's the responsibility of the user to figure our a different way. (though, as I said in #1 - I couldn't reproduce an unwanted "finally" execution)

slash/core/fixtures/active_fixture.py
```diff
@@ -33,10 +36,19 @@ def call_test_end(self):
...
+            return functools.partial(self.add_cleanup, success_only_cleanup=success_only_cleanup)
+
+        self._cleanups.append((cleanup, success_only_cleanup))
 
     def do_cleanups(self):
+        if not slash_context.result:
+            # for when we're unit-testing, and there's not 'result'
+            success = True
+        else:
+            success = slash_context.result.is_success()
```
koreno: @vmalloc - maybe there's a better way of determining `success`? it wasn't compatible with `test_fixture_scoping.py`

vmalloc:
Closing as this needs some some more brainstorming IMO

PR #1001, opened 2019-11-24
make 'TracebackProxy' compatible with 'inspect', which checks for types.TracebackType

koreno:
An attempt to use ipython's post-mortem code with `--pdb` failed deep inside, when ipython called on `inspect.getinnerframes`, which eventually failed on `inspect.getframeinfo`. The latter checks `if istraceback(frame)`, and falls into the wrong code path since `frame` is a TracebackProxy`

### DIFF
--- a/slash/utils/traceback_proxy.py
+++ b/slash/utils/traceback_proxy.py
@@ -148,6 +148,9 @@ else:
                 tb = sys.exc_info()[2]
             return tb
 
+        @property
+        def __class__(self):
+            return types.TracebackType
 
     def create_traceback_proxy(tb=None, frame_correction=0):
         """


slash/utils/traceback_proxy.py
```diff
@@ -176,3 +176,8 @@ def create_traceback_proxy(tb=None, frame_correction=0):
                 current = current.tb_next
 
         return (first, current)
+
+    def istraceback(object):
+        return isinstance(object, (types.TracebackType, TracebackProxy))
+
+    inspect.istraceback = istraceback  # so that it cooperates with 'inspect.getframeinfo'
```
vmalloc: avoid patching existing modules. You can achieve the same goal with a `__class__` property on TracebackProxy

koreno: you learn something new every day...:)

vmalloc:
Thanks!

PR #978, opened 2019-10-07
Add loglinkconverter plugin

bpowell65536:
When tests are run on a continuous integration system, this plugin converts the local filesystem links to log files given in the test session summary into web links into the artifact store for the job which ran the tests. As a result, the test summary sections for each failing test will contain a clickable link to the log file
for that test, making it easier to quickly diagnose a failing test.

PR #1003, opened 2019-12-24
Introduce end-of-line normalization

ayalash:
PR https://github.com/getslash/slash/pull/978 Added windows newline (`\n\r`) while the rest of the repository uses unix newline (`\n`).

I think we should normalized the repository to use only one of them.


ayalash:
@vmalloc What do you think?
If you agree with it, please merge.

PR #1024, opened 2020-06-17
1023: Support test filtering when tests are explicitly specified

yaelmi3:


slash/loader.py
```diff
@@ -154,6 +154,8 @@ def _iter_test_address(self, address):
                 if not self._address_in_file_matches(address_in_file, test):
                     continue
             matched = True
+            if self._is_excluded(test):
```
vmalloc, line 157: @yaelmi3 Isn't there a possibility that a test won't be excluded but one of its parametrizations will be? I think we used this capability in the past.

If I understand correctly this change will cause such tests to be run when they previously shouldn't have

yaelmi3: @vmalloc : Could you please elaborate, not sure I understand what I'm missing here.
In case we'd like to filter out one of test's params using "-k", it would behave in the following manner with this change:
```
@slash.tag("bla")
@slash.parameters.iterate(bla=[1, 2])
def test_bla1(bla):
    print(bla)


def test_bla2():
    pass
```
```
slash run tests/test_bla.py:test_bla1 tests/test_bla.py:test_bla2 -k "bla=1"
1 tests collected
tests/test_kaka.py:test_bla1(bla=1)
.  PASS
```

Whilst before it ended with :
```
slash run tests/test_kaka.py:test_bla1 tests/test_kaka.py:test_bla2 -k "bla=1"
ERROR: Cannot find test(s) for 'tests/test_kaka.py:test_bla2'
```

But that's not necessarily a bad thing IMO, since it was inconsistent with:
```
slash run tests/test_bla.py -k "bla=1" 
1 tests collected
 tests/test_bla.py:test_bla1(bla=1)
.  PASS
```
Or are you referring to `@slash.exclude`? From what I see this is enforced later in the runner (https://github.com/getslash/slash/blob/develop/slash/runner.py#L151)

vmalloc: Oh ok - as long as the filter you wrote works as expected I guess it's ok

vmalloc:
Thanks!

Issue #899, opened 2018-10-23
Handle slash.exclude decorator for fixtures


ayalash:
Currently, when a fixture is decorated with `slash.exclude`, slash silently ignores it. It should either exclude the parameters accordingly or forbid it.

ayalash:
After discussion with @vmalloc , it was decided to support `slash.exclude` for fixtures.

ayalash:
Easy reproduction, for this file:
```
$ cat /tmp/test_slash.py 
import slash


@slash.parameters.iterate(fx_param=['a', 'b'])
@slash.exclude("fx_param", ["a"])
@slash.fixture
def my_fixture(fx_param):
    pass

def test_fixutre(my_fixture):
    pass

@slash.parameters.iterate(fx_param=['a', 'b'])
@slash.exclude("fx_param", ["a"])
def test_without_fixture(fx_param):
    pass
```
Execution results:
```
$ slash run /tmp/test_slash.py
4 tests collected   
[2019-07-31 17:29:50] #1: /tmp/test_slash.py:test_fixutre(my_fixture.fx_param=a)
[2019-07-31 17:29:50] #2: /tmp/test_slash.py:test_fixutre(my_fixture.fx_param=b)
[2019-07-31 17:29:50] #3: /tmp/test_slash.py:test_without_fixture(fx_param=a)
Skipped: Excluded due to parameter combination exclusion rules, Test: </tmp/test_slash.py:test_without_fixture(fx_param=a)>
[2019-07-31 17:29:50] #4: /tmp/test_slash.py:test_without_fixture(fx_param=b)
======================================================= Session ended. 3 successful, 0 skipped, 0 failed, 0 erroneous. 1 not run. Total duration: 00:00:00 =======================================================
```

PR #1026, opened 2020-06-26
#899: Support slash.exclude decorator for fixtures

yaelmi3:


vmalloc:
@yaelmi3 I think this needs a changelog entry...

tests/test_fixture_exclusion.py
```diff
@@ -96,3 +96,26 @@ def test_invalid_exclude_patterns(names, values):
...
+    with slash.Session() as s:
+
+        @s.fixture_store.add_fixture
+        @slash.parametrize('param', [1, 2])
+        @slash.exclude("param", [1])
+        @slash.fixture
+        def fixture1(param):
+            return param
+
+        def test_something(fixture1):
```
vmalloc, line 112: I think this test case should have an additional fixture that is parametrized and not marked with exclude to make sure this plays well together...

yaelmi3: Fixed

vmalloc:
Thanks!

PR #1028, opened 2020-07-16
modified iter_suite_file_paths to get the test path as relative if the path is provided relative in the suite file

Rakesh-Dakshinamoorthy:
## Description
Slash creates test path and test address based on how it is invoked. when the test path provided is relative then the test path and address of a test will be relative. 
```
slash list tests\test_script.py
``` 
If the path is provided as absolute then the test_path and address will be absolute.
```
slash list C:\package\tests\test_script.py
```

But when slash list or slash run is invoked with help of suite file the test path and test address of the test collected is always absolute irrespective of path provided in suite file.

## Solution
modified the iter_suite_file_paths method to provide relative path when relative path is provided and absolute path when absolute path is provided in suite_file

### DIFF
--- a/slash/utils/suite_files.py
+++ b/slash/utils/suite_files.py
@@ -9,6 +9,7 @@ SuiteEntry = namedtuple('SuiteEntry', 'path, matcher, repeat')
 def iter_suite_file_paths(suite_files):
     for filename in suite_files:
 
+        dirname = os.path.abspath(os.path.dirname(filename))
         with open(filename) as suite_file:
             for path in suite_file:
                 path = path.strip()
@@ -17,6 +18,9 @@ def iter_suite_file_paths(suite_files):
 
                 suite_entry = _parse_path_filter_and_repeat(path)
                 path = suite_entry.path
+
+                if not os.path.isabs(path):
+                    path = os.path.relpath(os.path.join(dirname, path))
 
                 if not path.endswith('.py') and '.py:' not in path and not os.path.isdir(path):
                     for p, other_filter in iter_suite_file_paths([path]):
--- a/slash/utils/suite_files.py
+++ b/slash/utils/suite_files.py
@@ -9,7 +9,6 @@ SuiteEntry = namedtuple('SuiteEntry', 'path, matcher, repeat')
 def iter_suite_file_paths(suite_files):
     for filename in suite_files:
 
-        dirname = os.path.abspath(os.path.dirname(filename))
         with open(filename) as suite_file:
             for path in suite_file:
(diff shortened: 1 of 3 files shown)


slash/utils/suite_files.py
```diff
@@ -20,7 +20,7 @@ def iter_suite_file_paths(suite_files):
                 path = suite_entry.path
 
                 if not os.path.isabs(path):
-                    path = os.path.abspath(os.path.join(dirname, path))
+                    path = os.path.relpath(os.path.join(dirname, path))
```
vmalloc, line 23: I'm not sure I understand this change... On which cases is this `if` going to get entered, and have a different value for `path` afterwards?

If I pass `./bla`, relpath is going to still return `./bla`...

Rakesh-Dakshinamoorthy: The relative path in the suite-file will be relative path of the test script with respect to the suite-file path.
when the suite file path is in different path than the path where slash is invoked.
then path of test collected would be relative path of the script from where slash is invoked.
For example
When slash is invoked in same path as the sample_test_repo

image

```
slash run -f suites\module_1\suite_file.txt
```
or 
```
slash list -f suites\module_1\suite_file.txt
```
then the path of the tests collected will be relative to sample_test_repo.
tests\feature_one\test_sub_feature1.py

Rakesh-Dakshinamoorthy: @vmalloc do you have any more questions regarding this?

vmalloc: I see. @ayalash @yaelmi3 do you have any objection to this change?

yaelmi3: @vmalloc : LGTM

vmalloc:
@Rakesh-Dakshinamoorthy can you please include a changelog entry for this change? After this I will merge it in. Thanks and sorry for the delay!

Rakesh-Dakshinamoorthy:
> @Rakesh-Dakshinamoorthy can you please include a changelog entry for this change? After this I will merge it in. Thanks and sorry for the delay!

I have added the change log as you have requested

vmalloc:
@Rakesh-Dakshinamoorthy Thanks a lot for your contribution!

Rakesh-Dakshinamoorthy:
> @Rakesh-Dakshinamoorthy Thanks a lot for your contribution!

You are welcome @vmalloc.
I would like to know when this fix will be released? I would be really helpful for me.

vmalloc:
Yes. 1.11.0 will be out soon.

PR #1031, opened 2020-08-17
Fix tag param for class test

Rakesh-Dakshinamoorthy:
## Description
The New feature from slash 1.9.0 lets the tester to add tag to the test params. This tagging of test param is not working for class based test. Unable to filter the tests that are tagged in params of class based tests.

Sample code
```
class SampleTest(slash.Test):
    @slash.tag("settings")
    @slash.parametrize(
        "choose_door_keypad", ["new" // slash.param("T1", tags="bt-sanity"),
                               "existing" // slash.param("T2")])
    def test_5(self, choose_door_keypad):
        print("testing 5:" + choose_door_keypad)
``` 
In the above sample code tags are provided to param of class based test. when trying to filter the test with the tag. slash is unable to filter it.

```
slash run -k tag:bt-sanity <path-of-the-script>
```

## Solution
fetched the tags of params while getting the tags in get_tags method in class Test

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,5 +1,6 @@
 Changelog
 =========
+* :bug:`_` Fix to filter class based test based on the tagging in parameter
 * :bug:`899 major` Fix slash.exclude decorator for fixtures
 * :release:`1.10.0 <21-04-2020>`
 * :bug:`1013 major` Allow slash.use_fixtures application on other fixtures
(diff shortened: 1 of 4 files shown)


vmalloc:
@Rakesh-Dakshinamoorthy Good catch! Thanks for the contribution!

Rakesh-Dakshinamoorthy:
> hanks for the contribution!

You are welcome

PR #1049, opened 2021-01-10
Pin IPython<7.17.0 for Python3.6

### LINKED ISSUE #1048: slash doesn't work with python3.6 due to IPython's recent release
IPython recently (since version 7.17.0, released in Aug 2020) decided to drop support for Python>3.6. 

[Slash still supports Python3.6](https://github.com/getslash/slash/blob/develop/setup.cfg#L4).
 
A possible solution is to modify the requirements file so that if python==3.6, it will use IPython<7.17.0

Working on a pull request right now to do just that

d1618033:
Fixes #1048

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,5 +16,5 @@ pyparsing
 vintage>=0.4.0
 brotli
 IPython==1.2.1; implementation_name=='pypy'
-IPython<7.17.0; implementation_name!='pypy' and python_version=='3.6'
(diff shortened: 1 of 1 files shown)


d1618033:
The travis-ci is failing, but it doesn't look like it's related to my change. @vmalloc what should I do?

vmalloc:
@d1618033 I need to fix `develop`, but I will merge this PR anyways because it looks ok

PR #1052, opened 2022-02-08
issue-1051: Make parallel work with resume

### LINKED ISSUE #1051: Parallel doesn't work when running "slash resume" instead of "slash run"
The reason is that "run" is hardcoded in the WorkerConfiguration class

```python
# slash/parallel/worker_configuration.py
class WorkerConfiguration(object):
    def __init__(self, args, worker_id):
        self._worker_id = worker_id
        self.argv = [sys.executable, '-m', 'slash.frontend.main', 'run', '--parallel-parent-session-id', ctx.session.id] + \
                     args + ["--parallel-worker-id", str(worker_id)]
```
The problem is that args does not contain the cmd, so some refactoring is needed

d1618033:
Fixes #1051

### DIFF
--- a/slash/frontend/list_config.py
+++ b/slash/frontend/list_config.py
@@ -20,7 +20,7 @@ def _parse_args(args):
 
 def list_config(args, report_stream=sys.stdout):
 
-    args = _parse_args(args)
+    args = _parse_args(args.argv)
     _print = Printer(report_stream, force_color=args.force_color, enable_color=args.enable_color)
 
     filters = _parse_filters(args.paths)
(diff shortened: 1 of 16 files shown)


vmalloc (dismissed):
LGTM. Thanks!

vmalloc:
@d1618033 Unfortunately the CI no longer works and needs a transition to GH actions. I will merge this manually for now.

Thanks for the help!

vmalloc:
Oh sorry just noticed it is still marked WIP. Let me know when it's ready for review again.

vmalloc (approved):
LGTM

PR #1054, opened 2022-05-28
python 3.6 and ipython 7.17.0 are not compatible

dimbleby:
Presumably more or less no-one is using the exact python version 3.6 so this doesn't really matter, but you have a sort of out-by-one error here.


vmalloc (approved):

dimbleby:
hmm, not sure I'm right actually...

PR #1068, opened 2023-03-22
Add flag for resuming session without skipped tests

oren0e:
Adding `unstarted-only-without-skipped` option when resuming.

### DIFF
--- a/doc/slash_run.rst
+++ b/doc/slash_run.rst
@@ -109,7 +109,7 @@ When you run a session that fails, Slash automatically saves the tests intended 
 
   $ slash resume -vv <session id>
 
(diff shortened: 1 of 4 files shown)


oren0e:
I think branch protection rules need to be edited, see: https://github.com/orgs/community/discussions/30724#discussioncomment-3473663

oren0e:
@vmalloc @ayalash - can someone please review this?

vmalloc:
Why this and not just use "unstarted only" and "no skipped"?

oren0e:
You are correct. I changed it. Also added info to the docs

oren0e:
@vmalloc - Can you take a look please?

vmalloc (approved):

PR #1070, opened 2023-05-09
Remove xunit testcase classname attribute for test functions

pierreluctg:
Empty string` classname` is causing issues in some parsers and here the `classname` is really `None` and not `""`


vmalloc (approved):
LGTM

PR #1078, opened 2023-09-18
Simplify dependency specification for ipython

gragib:
I have had many issues with resolving dependencies in projects that use `poetry` or `pdm`. `ipython` 1.2.1 is used when installing slash in Python 3.7+ on Windows/Ubuntu/macOS. The changes proposed in this PR should preserve the current behaviour when using `pypy`, while allowing `cpython` implementations to use more recent versions of `ipython`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,9 +34,7 @@ dependencies = [
     "pyparsing",
     "vintage>=0.4.0",
     "brotli",
-    "IPython==1.2.1; implementation_name=='pypy'",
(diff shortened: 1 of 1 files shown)


vmalloc:
I don't think this change preserves the current behavior, as it unpins the dependency for pypy. Am I wrong?

gragib:
Moving this to draft while I gather the evidence.

gragib:
Is there any particular reason why `ipython` is pinned to `1.2.1` for `pypy`?

gragib:
Closing this since I can't even get the wheels on PyPI to install in pypy.

PR #1085, opened 2025-01-17
Fix the workflow to run on static version of Ubuntu

pierreluctg:
Change the runs-on from ubuntu-latest to ubuntu-22.04 to avoid picking up ubuntu-24.04 (current ubuntu-latest).

ubuntu-24.04 do not have Python 3.7 and is breaking the current matrix.

Future improvement would be to include the `runs-on` in the matrix itself.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pierreluctg:
FYI @ayalash

pierreluctg:
FYI @vmalloc

vmalloc (approved):
LGTM

